PR #4, opened 2025-04-16
Fix lint errors

manoelmarques:
Fix lint errors that fail CI
An empty Ginkgo test suite was added for now. Unit tests will be added in another PR.


.gitignore
```diff
@@ -1,2 +1,4 @@
 bin
 .vscode
+# MacOSX
+.DS_Store
```
elevran, line 4: nit: missing newline at end of file

pkg/vllm-sim/streaming.go
```diff
@@ -78,8 +78,16 @@ func (s *VllmSimulator) sendStreamingResponse(isChatCompletion bool, ctx *fastht
 		}
 
 		// finish sse events stream
-		fmt.Fprint(w, "data: [DONE]\n\n")
-		w.Flush()
+		_, err := fmt.Fprint(w, "data: [DONE]\n\n")
+		if err != nil {
+			ctx.Error("fprint failed, "+err.Error(), fasthttp.StatusInternalServerError)
+			return
```
elevran, line 84: there is a slight behavior change here - previously errors where ignored and `w.Flush()` and `s.responseSentCallback(model)` were both called.
This function now returns early - I don't know if there are side effects which may impact behavior

pkg/vllm-sim/streaming.go
```diff
@@ -130,8 +138,14 @@ func (s *VllmSimulator) sendChunk(isChatCompletion bool, w *bufio.Writer, creati
 		return err
 	}
 
-	fmt.Fprintf(w, "data: %s\n\n", data)
-	w.Flush()
+	_, err = fmt.Fprintf(w, "data: %s\n\n", data)
+	if err != nil {
+		return err
```
elevran, line 143: Potential behavior change - `w.Flush()` is no longer called and a non-nil error is returned to caller (does it expect/handle?)

PR #8, opened 2025-04-19
Add max_tokens/max_completion_tokens

manoelmarques:
Add chat completions request parameters max_tokens (deprecated) and max_completion_tokens
Add text completions request parameter max_tokens

Added unit tests for chat and text completions using those parameters.


pkg/vllm-sim/defs.go
```diff
@@ -190,20 +207,51 @@ type chatRespChunkChoice struct {
 }
 
 // createResponseText creates response text for the given chat completion request and mode
-func (req chatCompletionRequest) createResponseText(mode string) string {
+func (req chatCompletionRequest) createResponseText(mode string) (string, error) {
 	if mode == modeEcho {
-		return req.getLastUserMsg()
+		return req.getLastUserMsg(), nil
+	}
+	var typeToken string
```
mayabar, line 226: lines 214-231 and 241-252 are the same, please create a helper function

pkg/vllm-sim/defs.go
```diff
@@ -190,20 +207,51 @@ type chatRespChunkChoice struct {
 }
 
 // createResponseText creates response text for the given chat completion request and mode
-func (req chatCompletionRequest) createResponseText(mode string) string {
+func (req chatCompletionRequest) createResponseText(mode string) (string, error) {
 	if mode == modeEcho {
-		return req.getLastUserMsg()
+		return req.getLastUserMsg(), nil
```
mayabar: why max tokens parameters does not taken into consideration in echo mode?

pkg/vllm-sim/defs.go
```diff
@@ -190,20 +207,51 @@ type chatRespChunkChoice struct {
...
+	// use max_completion_tokens
+	// as in the real vllm
+	if req.MaxCompletionTokens != nil {
+		pointer = req.MaxCompletionTokens
+		typeToken = "max_completion_tokens"
+	} else if req.MaxTokens != nil {
+		pointer = req.MaxTokens
+		typeToken = "max_tokens"
+	}
+	maxTokens := int64(-1)
```
mayabar: I would use nil as a value with meaning "no limit"
If anyway -1 is preferred, please define a constant

pkg/vllm-sim/defs.go
manoelmarques: added helper function `getMaxTokens`

pkg/vllm-sim/defs.go
manoelmarques: use now max tokens on echo response

pkg/vllm-sim/defs.go
manoelmarques: changed to nil instead of -1

PR #15, opened 2025-05-14
update CONTAINER_TOOL in makefile

mayabar:


Makefile
```diff
@@ -13,9 +13,6 @@
 # limitations under the License.
 
 # Makefile for the vllm-sim project
-
```
elevran: did you mean to remove this? Is container *runtime* not used in Makefile?

PR #20, opened 2025-05-19
fix: Lint errors

shmuelk:
This PR corrects various issues pointed out by GolangCI-lint.

In addition the configuration for golangci-lint has been updated to allow dot-imports of the Ginkgo related packages. As they are typically used via dot-imports.

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -30,4 +30,10 @@ jobs:
       - name: Run go test
         shell: bash 
         run: |
+          echo "Installing Ginkgo..."
+          go install github.com/onsi/ginkgo/ginkgo@latest
+          export PATH=$PATH:$(go env GOPATH)/bin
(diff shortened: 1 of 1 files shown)


pkg/llm-d-inference-sim/defs.go
```diff
@@ -54,8 +54,8 @@ type VllmSimulator struct {
 	loraAdaptors sync.Map
 	// maxLoras defines maximum number of loaded loras
 	maxLoras int
-	// maxLoras defines maximum number of loras to store in CPU memory
-	maxCpuLoras int
+	// maxCPULoras defines maximum number of loras to store in CPU memory
+	maxCPULoras int
```
elevran, line 58: I would imagine this needs to be LoRAs as the acronym, but not critical and should not hold the PR

.golangci.yml
elevran, line 1: consider adding reporting on new code only - should speed up processing if it becomes an issue

PR #21, opened 2025-05-21
link checker - fails PR if links are broken

clubanderson:


shmuelk (approved):
/lgtm

/approve

elevran:
See llm-d/llm-d-inference-scheduler#130 for what we settled on (e.g., workflow vs action, different checker)

shmuelk:
We have decided to align our tooling within llm-d as much as possible. As such we want to align the llm-d-inference-sim's broken link checker with that used in the llm-d-inference-scheduler.

This PR is being closed and another one will be created.

Issue #29, opened 2025-05-29
Send finish_reason length if the response is cut because of max_tokens


irar2:
If the response is cut because of max_tokens, finish_reason should be length. At the moment, the only finish_reason that is sent is stop.

In addition, in case of streaming, the last chunk should contain the last part of the response.

irar2:
Implemented in #33

PR #33, opened 2025-05-29
Set correct finish reason

irar2:
#29 

Send finish reason length if the response text was truncated because of max_tokens.
In case of streaming and finish reason length, send finish reason with the last part of the text and not in a separate chunk


shmuelk (approved):
/lgtm

/approve

PR #36, opened 2025-05-29
Updates Metrics Reference on Readme

danehans:
The Readme currently states `vllm:lora_requests_info` is the only supported vLLM Prometheus metric. This PR updates the Readme to reference the metrics source code instead of maintaining a static list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@ Currently it supports partial OpenAI-compatible API:
 - /v1/completions 
 - /v1/models
 
-In addition, it supports a subset of vLLM's Prometheus metrics. These metrics are exposed via the /metrics HTTP REST endpoint. Currently supported are the following metrics:
-- vllm:lora_requests_info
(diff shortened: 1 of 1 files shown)


myf5:
Seem like the simulator does not support for `vllm:gpu_cache_usage_perc`,`vllm:num_requests_running`,`vllm:num_requests_waiting` . I tested on my Macbookpro,  found these 3 metrics keep 0.

Can we support these 3 metrics, it will be helpful to test llm workload aware loadbalancing.

mayabar:
Hi @myf5 
`vllm:gpu_cache_usage_perc` reports contant value = 0
`vllm:num_requests_waiting` returns real number of requests in the waiting queue. You can use command line parameter `max-running-requests` to define the maximum number of request that are processed in parallel. In addition you can use `time-to-first-token` command line parameter to define request processing time. Once you will run with this configuration, and send enough requests you will get value of `vllm:num_requests_waiting` greater than 0

Issue #31, opened 2025-05-29
Add object field  to response


irar2:
Example,
data: {"id":"...","object":"chat.completion.chunk","created":..,"model":"...","choices":[...]}

senanz:
Hi @irar2 ,
For all API requests ?

- /v1/chat/completions
- /v1/completions
- /v1/models

or specific one?

PR #47, opened 2025-06-04
Add object field to responses

irar2:
Implementation of #31

### DIFF
--- a/pkg/llm-d-inference-sim/defs.go
+++ b/pkg/llm-d-inference-sim/defs.go
@@ -29,16 +29,16 @@ import (
 )
 
 const (
-	modeRandom          = "random"
-	modeEcho            = "echo"
-	chatComplIDPrefix   = "chatcmpl-"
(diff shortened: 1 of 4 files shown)


pkg/llm-d-inference-sim/defs.go
```diff
@@ -29,13 +29,16 @@ import (
...
-	roleAssistant      = "assistant"
-	roleUser           = "user"
+	modeRandom          = "random"
+	modeEcho            = "echo"
+	chatComplIDPrefix   = "chatcmpl-"
+	stopFinishReason    = "stop"
+	lengthFinishReason  = "length"
+	roleAssistant       = "assistant"
+	roleUser            = "user"
+	textCompletion      = "text_completion"
```
mayabar: Maybe add "object" to names, textCompletion is too general

irar2: Done

mayabar (approved):

Issue #48, opened 2025-06-04
Multimodal Requests Not Supported


JuanmaBM:
Currently, the simulator does not support multimodal request payloads, as used by frameworks like Llama Stack or any OpenAI-compatible client that sends structured content blocks (e.g., text, image_url, video_url).

These clients send a content array of objects instead of a plain string. When such a request is received, the simulator fails to parse the body, expecting content to be of type string, not []object.

### 🔁 Example Request (Valid OpenAI format)

```
curl -k -X POST http://localhost:8000/v1/chat/completions \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "inference-simulator-maas",
    "messages": [
      {
        "role": "user",
        "content": [
          { "type": "text", "text": "Can you describe this image?" },
          { "type": "image_url", "url": "https://example.com/image.png" }
        ]
      }
    ]
  }'
```

### ❌ Current Result

```
E0604 09:16:14.284161       1 simulator.go:264] "failed to read and parse request body" err="json: cannot unmarshal array into Go struct field message.messages.content of type string"
```

### ✅ Expected Behavior
The simulator should:

* Accept content as an array of typed blocks (type: "text", type: "image_url", etc.)
* Gracefully handle or skip unsupported types (e.g., images or videos) if not yet implemented
* Provide fallback or mock behavior where needed for inference simulation

PR #49, opened 2025-06-04
add support to multimodal in chat completions

JuanmaBM:
This PR adds support for OpenAI-style multimodal request formats, allowing the simulator to correctly parse messages where the content field is an array of typed blocks (text, image_url, etc.) instead of a plain string.

### Key Changes:
* Refactors the Message struct to support []ContentBlock instead of string.
* Adds parsing logic.
* Maintains backward compatibility with simple string content if needed.

This enables compatibility with frameworks like Llama Stack and other OpenAI v2 clients using structured content.

Fixes: [Multimodal Requests Not Supported](https://github.com/llm-d/llm-d-inference-sim/issues/48)

### DIFF
--- a/pkg/llm-d-inference-sim/defs.go
+++ b/pkg/llm-d-inference-sim/defs.go
@@ -227,6 +227,7 @@ func (mc content) PlainText() string {
 	for _, block := range mc.Structured {
 		if block.Type == "text" {
 			sb.WriteString(block.Text)
+			sb.WriteString(" ")
 		}
 	}
 	return sb.String()


mayabar:
Hi @JuanmaBM , thanks for you PR.
Support of multimodal requests in chat completion is an important feature.
According the OpenAI API (https://platform.openai.com/docs/guides/images-vision?api-mode=chat) structure of content in case of array of items seems a little different from one in the implementation.
OpenAI API's example: 
```
"content": [
            {"type": "text", "text": "What's in this image?"},
            {
                "type": "image_url",
                "image_url": {
                    "url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg",
                },
            },
        ],
```
Code:
```
type contentBlock struct {
	Type     string `json:"type"`
	Text     string `json:"text,omitempty"`
	ImageURL string `json:"image_url,omitempty"`
}
```
image_url is string instead of object with 'url' field.

mayabar:
Thanks for your comment, @JuanmaBM , in the simulator we want to be compatible with OpenAI API specification, at this stage we don't support all fields in requests data structures, currently, our goal is to support a sub-set of the OpenAI API, but we don't want to add data structure's fields that are not defined in OpenAI API or overload existing fields. This will allow the simulator clients to be able easily to switch to real vLLM

mayabar:
Hi @JuanmaBM , thank you for the update.

Regarding the schema validation - the request body structure is validated by marshaling to an appropriate structure.
Ira (@irar2) is working on another PR that will add support for tools calls. Tool description is a "free" json that should be compliant to the specific schema, in this case we will use schema validation.

pkg/llm-d-inference-sim/defs.go
```diff
@@ -170,7 +172,64 @@ type message struct {
...
+}
+
+func (mc content) PlainText() string {
+	if mc.Raw != "" {
+		return mc.Raw
+	}
+	var sb strings.Builder
+	for _, block := range mc.Structured {
+		if block.Type == "text" {
+			sb.WriteString(block.Text)
```
mayabar, line 229: maybe add here separator (space), this will allow better tokens number calculation

JuanmaBM: Done

PR #61, opened 2025-06-18
Calculate usage data once for both text and chat

### LINKED ISSUE #56: Code refactoring
Split defs.go into three files: create separate files for requests and responses.
Calculate usage in one place for both streaming and non-streaming

irar2:
We now calculate usage data in one place for both text and chat completion, and pass it to the relevant functions.

This is the second part of #56 (fixes #56)

### DIFF
--- a/pkg/llm-d-inference-sim/response.go
+++ b/pkg/llm-d-inference-sim/response.go
@@ -20,7 +20,6 @@ package llmdinferencesim
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/llm-d-inference-sim/response.go
```diff
@@ -283,12 +284,13 @@ func (req chatCompletionRequest) createToolCalls() ([]toolCall, string, error) {
 		}
 		calls = append(calls, call)
 	}
-	return calls, toolsFinishReason, nil
+
+	return calls, toolsFinishReason, countTokensForToolCalls(calls), nil
```
shmuelk: Since in the case of without tool calls you calculate the number of tokens in the calling function and not in createResponseText, I think you should do the same here. The call to countTokensForToolsCalls should be in the calling function.

Also, I don't see the definition of countTokensForToolCalls which I assume is new code

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -558,6 +561,7 @@ func (s *VllmSimulator) sendResponse(isChatCompletion bool, ctx *fasthttp.Reques
 	}
 
 	// calculate how long to wait before returning the response, time is based on number of tokens (use words number)
+	numOfWords := usageData.PromptTokens
```
mayabar, line 564: do we need numOfWords variable?

mayabar (changes requested):

pkg/llm-d-inference-sim/response.go
irar2: Moving the calculation to the caller will add another if statement there. I think it's more clear my way, we calculate the tokens if we generate the calls

pkg/llm-d-inference-sim/simulator.go
irar2: I think it makes the code in the next line more readable. Also, this is how the code was before adding the usage.

mayabar (approved):

shmuelk (approved):

PR #62, opened 2025-06-22
Improved tokenization

### LINKED ISSUE #51: Improve tokens creations
At the moment, we split response text into tokens by spaces. It may be useful to split also by other delimiters.

irar2:
At the moment, we split the response text (and generated tool arguments) to tokens by spaces. This PR implements tokenization by special characters as well.
Fixes #51


pkg/llm-d-inference-sim/utils.go
```diff
@@ -124,3 +126,8 @@ func randomInt(max int, startFromOne bool) int {
 func flipCoin() bool {
 	return randomInt(1, false) != 0
 }
+
+func tokenize(text string) []string {
+	re := regexp.MustCompile(`(\{|\}|:|,|-|\.|\?|\!|;|@|#|\$|%|\^|&|\*|\(|\)|\+|\-|_|~|/|\\|>|<|\[|\]|=|"|\w+)(\s*)`)
```
shmuelk: Compile the regular expression once in an init() function. MustCompile panics if the compile fails.

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -511,11 +511,14 @@ func (s *VllmSimulator) HandleError(_ *fasthttp.RequestCtx, err error) {
 	s.logger.Error(err, "VLLM server error")
 }
 
-// createCompletionResponse creates response for completion request, supports both completion requests types (text and chat) as defined by isChatCompletion
-// respText - content to be sent in the response
+// createCompletionResponse creates response for completion request, supports both completion requests types (text and chat)
```
shmuelk: Change text to:

createCompletionResponse creates the response for completion requests, supports both completion request types (text and chat)

pkg/llm-d-inference-sim/streaming.go
```diff
@@ -34,20 +33,19 @@ type streamingContext struct {
 	creationTime     int64
 }
 
-// sendStreamingResponse creates and sends a streaming response for completion request of both types (text and chat) as defined by isChatCompletion
+// sendStreamingResponse creates and sends a streaming response for completion request of both types (text and chat)
```
shmuelk: Change the text to:

sendStreamingResponse creates and sends a streaming response for completion requests of both types (text and chat)

shmuelk (changes requested):

pkg/llm-d-inference-sim/utils.go
irar2: done

pkg/llm-d-inference-sim/simulator.go
irar2: done

pkg/llm-d-inference-sim/streaming.go
irar2: done

pkg/llm-d-inference-sim/request.go
```diff
@@ -154,11 +155,13 @@ func (req *chatCompletionRequest) getLastUserMsg() string {
 	return ""
 }
 
-// createResponseText creates response text for the given chat completion request and mode
-func (req chatCompletionRequest) createResponseText(mode string) (string, string, error) {
+// createResponseText creates and returns response payload based on this request,
+// i.e., an array of generated tokens, the finish reason, and the number of created
+// tokens
+func (req chatCompletionRequest) createResponseText(mode string) ([]string, string, int, error) {
```
mayabar, line 161: In the current version array of created tokens is returned, why we need the number of created tokens as an additional parameter? it's always len(tokens), isn't it?

pkg/llm-d-inference-sim/request.go
```diff
@@ -198,11 +202,13 @@ func (c *textCompletionRequest) getToolChoice() string {
 	return ""
 }
 
-// createResponseText creates response text for the given text completion request and mode
-func (req textCompletionRequest) createResponseText(mode string) (string, string, error) {
+// createResponseText creates and returns response payload based on this request,
+// i.e., an array of generated tokens, the finish reason, and the number of created
+// tokens
+func (req textCompletionRequest) createResponseText(mode string) ([]string, string, int, error) {
```
mayabar, line 208: same question regarding the returned int value

pkg/llm-d-inference-sim/request.go
irar2: Right, this is just to return similar values to createToolCalls

PR #78, opened 2025-07-06
Support space separated arguments, use correct format in config file

irar2:
This PR allows space separated command line arguments (served-model-name and lora-modules), and fixes the example config file format


irar2:
Thanks, there are tests for these cases for `--served-model-name`

PR #81, opened 2025-07-09
use string when storing lora adapters in simulator

mayabar:


pkg/llm-d-inference-sim/lora.go
```diff
@@ -33,12 +33,13 @@ type unloadLoraRequest struct {
...
 
 func (s *VllmSimulator) getLoras() []string {
-
 	loras := make([]string, 0)
 
 	s.loraAdaptors.Range(func(key, _ any) bool {
 		if lora, ok := key.(string); ok {
 			loras = append(loras, lora)
+		} else {
+			s.logger.Info("Stored LoRA is not a string", "value", lora)
```
irar2: I think you want to print the value of key in case of an error, because lora will be nil

Issue #82, opened 2025-07-10
Support --max-model-len config parameter


mayabar:
`max-model-len` parameter defines model's context window, number of tokens in a single request including input and output. Default value is 1024.
Request that does not aligned with defined values will fail with error code 400 (bad request)

**Example request:**
```
{
  prompt: "string of length X",
  max_tokens: Y
}
```
max-model-len=Z

Request will return error if X+Y > Z
HTTP status = 400, bad request

**Payload example:**
```
{
  "object":"error",
  "message":"This model's maximum context length is <Z> tokens. However, you requested <Y> tokens (<X> in the messages, <Y> in the completion). Please reduce the length of the messages or completion.",
  "type":"BadRequestError",
  "param":null,
  "code":400
}
```

PR #85, opened 2025-07-12
feat: add max-model-len configuration and validation for context window (#82)

mohitpalsingh:
# max-model-len Feature Implementation

## Overview

This implementation adds support for the `max-model-len` parameter, which defines the model's context window - the maximum number of tokens in a single request including both input and output tokens.

### Configuration
- **Parameter**: `--max-model-len`
- **Default Value**: 1024 tokens
- **Type**: Integer
- **Description**: Defines the model's context window size

### Validation Logic
When a request is received, the simulator validates:
```
prompt_tokens + max_completion_tokens <= max_model_len
```

If this condition is violated, the request is rejected with HTTP 400 status code.

### Error Response Format
When the context window limit is exceeded, the following error response is returned:

```json
{
  "object": "error",
  "message": "This model's maximum context length is <Z> tokens. However, you requested <Y> tokens (<X> in the messages, <Y> in the completion). Please reduce the length of the messages or completion.",
  "type": "BadRequestError",
  "param": null,
  "code": 400
}
```

Where:
- `<Z>` = max_model_len value
- `<X>` = number of tokens in the prompt/messages
- `<Y>` = max_tokens requested for completion

### Files Modified

1. **config.go**
   - Added `MaxModelLen int` field to configuration struct
   - Added default value (1024) in `newConfig()`
   - Added validation in `validate()` method

2. **simulator.go**
   - Added command line flag `--max-model-len`
   - Added context window validation in `handleCompletions()`
   - Fixed `sendCompletionError()` to use correct HTTP status code

3. **utils.go**
   - Added `validateContextWindow()` function for validation logic

4. **request.go**
   - Added `getMaxCompletionTokens()` method to completionRequest interface
   - Implemented the method for both chat and text completion requests

### Test Coverage

Added comprehensive tests covering:
- Configuration validation for invalid max-model-len values
- Context window validation logic (unit tests)
- Integration tests for both chat and text completion APIs
- HTTP response format validation
- Successful requests within context window limits

### Command Line
```bash
# Start simulator with 2048 token context window
./llm-d-inference-sim --model my-model --max-model-len 2048

# Default 1024 token context window
./llm-d-inference-sim --model my-model
```

### Configuration File
```yaml
max-model-len: 2048
```

### API Requests

**Request that exceeds context window:**
```bash
curl -X POST http://localhost:8000/v1/chat/completions \
  -d '{
    "model": "my-model",
    "messages": [{"role": "user", "content": "This is a very long prompt..."}],
    "max_tokens": 100
  }'
```

**Response (HTTP 400):**
```json
{
  "object": "error",
  "message": "This model's maximum context length is 50 tokens. However, you requested 120 tokens (20 in the messages, 100 in the completion). Please reduce the length of the messages or completion.",
  "type": "BadRequestError",
  "param": null,
  "code": 400
}
```

## Compatibility

- Works with both chat completion (`/v1/chat/completions`) and text completion (`/v1/completions`) endpoints
- Supports both `max_tokens` and `max_completion_tokens` parameters
- Compatible with streaming and non-streaming requests
- Maintains backward compatibility (feature is optional with sensible default)

## Testing

Run the test suite to verify the implementation:
```bash
go test ./pkg/llm-d-inference-sim/ -v
```

All tests pass (88/88 specs) including the new context window validation tests.

### DIFF
--- a/pkg/llm-d-inference-sim/config_test.go
+++ b/pkg/llm-d-inference-sim/config_test.go
@@ -194,6 +194,12 @@ var _ = Describe("Simulator configuration", func() {
 	}
 	tests = append(tests, test)
 
+	test = testCase{
+		name: "invalid max-model-len",
+		args: []string{"cmd", "--max-model-len", "0", "--config", "../../manifests/config.yaml"},
+	}
+	tests = append(tests, test)
+
 	DescribeTable("check configurations",
 		func(args []string, expectedConfig *configuration) {
 			config, err := createSimConfig(args)
@@ -217,5 +223,6 @@ var _ = Describe("Simulator configuration", func() {
 		Entry(tests[7].name, tests[7].args),
 		Entry(tests[8].name, tests[8].args),
 		Entry(tests[9].name, tests[9].args),
+		Entry(tests[10].name, tests[10].args),
 	)
 })
--- a/pkg/llm-d-inference-sim/utils.go
+++ b/pkg/llm-d-inference-sim/utils.go
@@ -58,6 +58,27 @@ func getMaxTokens(maxCompletionTokens *int64, maxTokens *int64) (*int64, error) 
 	return tokens, nil
 }
 
+// validateContextWindow checks if the request fits within the model's context window
+// Returns an error if prompt tokens + max completion tokens exceeds the max model length
+func validateContextWindow(promptTokens int, maxCompletionTokens *int64, maxModelLen int) error {
+	if maxModelLen <= 0 {
+		return nil // no limit configured
+	}
+
+	completionTokens := int64(0)
+	if maxCompletionTokens != nil {
+		completionTokens = *maxCompletionTokens
+	}
+
+	totalTokens := int64(promptTokens) + completionTokens
+	if totalTokens > int64(maxModelLen) {
+		return fmt.Errorf("This model's maximum context length is %d tokens. However, you requested %d tokens (%d in the messages, %d in the completion). Please reduce the length of the messages or completion.",
+			maxModelLen, totalTokens, promptTokens, completionTokens)
+	}
+
+	return nil
+}
+
 // getRandomResponseText returns random response text from the pre-defined list of responses
 // considering max completion tokens if it is not nil, and a finish reason (stop or length)
 func getRandomResponseText(maxCompletionTokens *int64) (string, string) {
--- a/pkg/llm-d-inference-sim/utils_test.go
+++ b/pkg/llm-d-inference-sim/utils_test.go
@@ -43,4 +43,54 @@ var _ = Describe("Utils", func() {
 			Expect(finishReason).Should(Equal(stopFinishReason))
 		})
 	})
+
+	Context("validateContextWindow", func() {
+		It("should pass when total tokens are within limit", func() {
+			promptTokens := 100
+			maxCompletionTokens := int64(50)
+			maxModelLen := 200
+
+			err := validateContextWindow(promptTokens, &maxCompletionTokens, maxModelLen)
+			Expect(err).Should(BeNil())
+		})
+
+		It("should fail when total tokens exceed limit", func() {
+			promptTokens := 150
+			maxCompletionTokens := int64(100)
+			maxModelLen := 200
+
+			err := validateContextWindow(promptTokens, &maxCompletionTokens, maxModelLen)
+			Expect(err).ShouldNot(BeNil())
+			Expect(err.Error()).Should(ContainSubstring("This model's maximum context length is 200 tokens"))
+			Expect(err.Error()).Should(ContainSubstring("However, you requested 250 tokens"))
+			Expect(err.Error()).Should(ContainSubstring("150 in the messages, 100 in the completion"))
+		})
+
+		It("should pass when no max model length is configured", func() {
+			promptTokens := 1000
+			maxCompletionTokens := int64(1000)
+			maxModelLen := 0
+
+			err := validateContextWindow(promptTokens, &maxCompletionTokens, maxModelLen)
+			Expect(err).Should(BeNil())
+		})
+
+		It("should handle nil max completion tokens", func() {
+			promptTokens := 100
+			maxModelLen := 200
+
+			err := validateContextWindow(promptTokens, nil, maxModelLen)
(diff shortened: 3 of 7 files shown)


pkg/llm-d-inference-sim/utils.go
```diff
@@ -58,6 +58,27 @@ func getMaxTokens(maxCompletionTokens *int64, maxTokens *int64) (*int64, error)
 	return tokens, nil
 }
 
+// validateContextWindow checks if the request fits within the model's context window
+// Returns an error if prompt tokens + max completion tokens exceeds the max model length
+func validateContextWindow(promptTokens int, maxCompletionTokens *int64, maxModelLen int) error {
+	if maxModelLen <= 0 {
+		return nil // no limit configured
+	}
```
irar2: Is it possible that maxModelLen is <= 0? You added a check in configuration validate()
`if c.MaxModelLen < 1 {
		return errors.New("max model len cannot be less than 1")
	}`

pkg/llm-d-inference-sim/utils_test.go
```diff
@@ -43,4 +43,54 @@ var _ = Describe("Utils", func() {
...
+			Expect(err).ShouldNot(BeNil())
+			Expect(err.Error()).Should(ContainSubstring("This model's maximum context length is 200 tokens"))
+			Expect(err.Error()).Should(ContainSubstring("However, you requested 250 tokens"))
+			Expect(err.Error()).Should(ContainSubstring("150 in the messages, 100 in the completion"))
+		})
+
+		It("should pass when no max model length is configured", func() {
+			promptTokens := 1000
+			maxCompletionTokens := int64(1000)
+			maxModelLen := 0
```
irar2: Likewise

irar2 (commented):
Thanks a lot for the PR! Very thorough. Please see a minor comment

pkg/llm-d-inference-sim/utils.go
mohitpalsingh: Hey @irar2 , you're right. This is redundant. Removing this.

pkg/llm-d-inference-sim/utils_test.go
mohitpalsingh: Removing this.

pkg/llm-d-inference-sim/utils.go
mohitpalsingh: Done, Can you please check now?

pkg/llm-d-inference-sim/config_test.go
```diff
@@ -258,6 +264,6 @@ var _ = Describe("Simulator configuration", func() {
 		Entry(tests[8].name, tests[8].args),
 		Entry(tests[9].name, tests[9].args),
 		Entry(tests[10].name, tests[10].args),
-		Entry(tests[11].name, tests[11].args),
+    Entry(tests[11].name, tests[11].args),
```
irar2: The merge is incorrect, there are 11 tests in main, and you added another one, so there should be an additional line with 
`Entry(tests[12].name, tests[12].args)`

mohitpalsingh: Missed this one, thanks.

mohitpalsingh: resolved

pkg/llm-d-inference-sim/utils.go
```diff
@@ -58,6 +58,23 @@ func getMaxTokens(maxCompletionTokens *int64, maxTokens *int64) (*int64, error)
...
+// Returns an error if prompt tokens + max completion tokens exceeds the max model length
+func validateContextWindow(promptTokens int, maxCompletionTokens *int64, maxModelLen int) error {
+	completionTokens := int64(0)
+	if maxCompletionTokens != nil {
+		completionTokens = *maxCompletionTokens
+	}
+
+	totalTokens := int64(promptTokens) + completionTokens
+	if totalTokens > int64(maxModelLen) {
+		return fmt.Errorf("this model's maximum context length is %d tokens. However, you requested %d tokens (%d in the messages, %d in the completion). Please reduce the length of the messages or completion",
```
mayabar: We want to be consistent with vLLM's error messages, which start with capital letter.
Please return error message to be sent to the client and use it in HTTP response creation.

mohitpalsingh:
is there any other blocker to this @irar2 ?
cc: @mayabar

pkg/llm-d-inference-sim/utils.go
mohitpalsingh: I changed that to lowercase to eliminate a static error check but I guess that was only for error messages ending with punctuation. Nevertheless I fixed it now.

Let me know if there is any other issue?

pkg/llm-d-inference-sim/utils.go
```diff
@@ -58,6 +58,23 @@ func getMaxTokens(maxCompletionTokens *int64, maxTokens *int64) (*int64, error)
...
+// Returns an error if prompt tokens + max completion tokens exceeds the max model length
+func validateContextWindow(promptTokens int, maxCompletionTokens *int64, maxModelLen int) error {
+	completionTokens := int64(0)
+	if maxCompletionTokens != nil {
+		completionTokens = *maxCompletionTokens
+	}
+
+	totalTokens := int64(promptTokens) + completionTokens
+	if totalTokens > int64(maxModelLen) {
+		return fmt.Errorf("This model's maximum context length is %d tokens. However, you requested %d tokens (%d in the messages, %d in the completion). Please reduce the length of the messages or completion",
```
mayabar: Hi @mohitpalsingh, the error message cannot be capitalized (the error: "error strings should not be capitalized"), and actually there is no need in error, we need only message that should be returned in HTTP response. This is the reason I suggested to return error message only and use it in the HTTP response payload.
You can run locally "make lint" to test lint issues.

mayabar (changes requested):

pkg/llm-d-inference-sim/utils.go
mohitpalsingh: @mayabar I got the issue now. I've refactored the validation function to instead return bool and form the error inside the caller function. This solved both lint and coherency issues.

Let me know if this approach works?

mayabar (approved):

Issue #83, opened 2025-07-10
Support `--max-num-batched-tokens` configuration parameter


mayabar:
**Definition**: `max-num-batched-tokens` is maximum number of batched tokens per iteration.

Currently only `max-num-seqs` is supported and defines how many requests can run in parallel.

If `max-num-batched-tokens` is defined the actual total number of tokens allowed in a batch at runtime should be limited.

Given a queue of received requests, the worker should pull a request from the queue only if both of the following conditions are met:

- The number of currently running requests is less than `max-num-seqs`, and

- The total number of processing tokens is less than `max-num-batched-tokens`.

The total number of processing tokens is calculated by summing, for each running request, the number of tokens in the prompt plus the maximum number of tokens in the output.
The number of output tokens is defined by the `max_tokens` parameter in the request. If `max_tokens` is not specified, it is calculated by  `max-model-len - prompt-len`.

mohitpalsingh:
Hey @mayabar , is someone already working on this?

I can pick this up if that's okay by you.

mayabar:
Sure, you are welcome to work on this ;)

PR #97, opened 2025-07-15
feat: add max-num-batched-tokens configuration and implement request handling constraints (#83)

mohitpalsingh:
## ✨ New Feature: `max-num-batched-tokens` Support

This PR implements the `max-num-batched-tokens` parameter, which limits the total number of tokens (prompt + max output tokens) that can be processed simultaneously across all running requests.

### 🔧 Technical Implementation
* Added `calculateProcessingTokens()` function to compute total token requirements per request.
* Implemented `canAcceptRequest()` function to check constraint satisfaction.
* Added `addRunningRequest()` and `removeRunningRequest()` for proper token tracking.
* Redesigned queue manager with ticker-based processing to avoid busy-waiting.
* Added atomic counters for thread-safe token counting.

### 📝 Configuration & Documentation
* Added command-line parameter: `--max-num-batched-tokens` with proper help text.
* Updated configuration files: Added parameter to `config.yaml` (2048) and `basic-config.yaml` (1024).
* Enhanced README: Added documentation for the new parameter.
* Updated manifests: Configuration examples now include the new parameter.

### 🔄 Code Quality Improvements
* Maintained backward compatibility - when set to `0`, only the `max-num-seqs` constraint is enforced.

---

### 🧪 Testing
* ✅ All existing tests pass (102/102 specs).
* ✅ Added comprehensive test coverage for the new functionality.
* ✅ Test scenarios include:
    * Parameter validation and edge cases
    * Queue constraint enforcement
    * Token calculation accuracy
    * Configuration loading from files
    * Backward compatibility (disabled mode)

---

### 📊 Behavior

**When `max-num-batched-tokens` is configured:**
* Requests are queued until both constraints are satisfied:
    * Number of running requests < `max-num-seqs`
    * Total processing tokens + new request tokens ≤ `max-num-batched-tokens`

**When `max-num-batched-tokens` is `0` or not set:**
* Only the `max-num-seqs` constraint is enforced (existing behavior).

---

### 🏗️ Files Modified
* `config.go` - Added parameter and validation
* `simulator.go` - Core implementation and renamed field
* `pkg/llm-d-inference-sim/*_test.go` - Updated tests and expectations
* `config.yaml` & `basic-config.yaml` - Added parameter
* `README.md` - Documentation updates

---

### ✅ Validation
- [x] Build succeeds without errors
- [x] All tests pass (102/102 specs)
- [x] Parameter appears in `--help` output
- [x] Configuration loading works correctly
- [x] Token tracking is thread-safe
- [x] Backward compatibility maintained

---

### 🎯 Addresses
This implementation follows the `vLLM` specification for the `max-num-batched-tokens` parameter, ensuring requests only proceed when both sequence and token constraints are satisfied. This enables better resource management and throughput control.

### DIFF
--- a/pkg/llm-d-inference-sim/config_test.go
+++ b/pkg/llm-d-inference-sim/config_test.go
@@ -53,6 +53,7 @@ func createDefaultConfig(model string) *configuration {
 	c.MaxNumSeqs = 5
 	c.MaxLoras = 2
 	c.MaxCPULoras = 5
+	c.MaxNumBatchedTokens = 2048
 	c.TimeToFirstToken = 2000
 	c.InterTokenLatency = 1000
 	c.KVCacheTransferLatency = 100
@@ -88,7 +89,6 @@ var _ = Describe("Simulator configuration", func() {
 	c = createDefaultConfig(qwenModelName)
 	c.Port = 8001
 	c.ServedModelNames = []string{"model1", "model2"}
-	c.MaxNumBatchedTokens = 2048
 	c.LoraModules = []loraModule{{Name: "lora1", Path: "/path/to/lora1"}, {Name: "lora2", Path: "/path/to/lora2"}}
 	test = testCase{
 		name:           "config file",
@@ -106,7 +106,6 @@ var _ = Describe("Simulator configuration", func() {
 	c.Port = 8002
 	c.ServedModelNames = []string{"alias1", "alias2"}
 	c.Seed = 100
-	c.MaxNumBatchedTokens = 2048
 	c.LoraModules = []loraModule{{Name: "lora3", Path: "/path/to/lora3"}, {Name: "lora4", Path: "/path/to/lora4"}}
 	c.LoraModulesString = []string{
 		"{\"name\":\"lora3\",\"path\":\"/path/to/lora3\"}",
@@ -125,7 +124,6 @@ var _ = Describe("Simulator configuration", func() {
 	// Config from config.yaml file plus command line args with different format
 	c = createDefaultConfig(model)
 	c.Port = 8002
-	c.MaxNumBatchedTokens = 2048
 	c.LoraModules = []loraModule{{Name: "lora3", Path: "/path/to/lora3"}}
 	c.LoraModulesString = []string{
 		"{\"name\":\"lora3\",\"path\":\"/path/to/lora3\"}",
@@ -143,7 +141,6 @@ var _ = Describe("Simulator configuration", func() {
 	// Config from config.yaml file plus command line args with empty string
 	c = createDefaultConfig(model)
 	c.Port = 8002
-	c.MaxNumBatchedTokens = 2048
 	c.LoraModules = []loraModule{{Name: "lora3", Path: "/path/to/lora3"}}
 	c.LoraModulesString = []string{
 		"{\"name\":\"lora3\",\"path\":\"/path/to/lora3\"}",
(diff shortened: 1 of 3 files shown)


pkg/llm-d-inference-sim/config_test.go
```diff
@@ -88,6 +88,7 @@ var _ = Describe("Simulator configuration", func() {
 	c = createDefaultConfig(qwenModelName)
 	c.Port = 8001
 	c.ServedModelNames = []string{"model1", "model2"}
+	c.MaxNumBatchedTokens = 2048
```
irar2: Please move this (and all other occurrences) to createDefaultConfig().

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -375,6 +393,72 @@ func (s *VllmSimulator) isLora(model string) bool {
...
+	// If max_tokens is not specified, calculate it as max-model-len - prompt-len
+	outputTokens := 0
+	if maxCompletionTokens != nil {
+		outputTokens = int(*maxCompletionTokens)
+	} else {
+		outputTokens = s.config.MaxModelLen - promptTokens
+		if outputTokens < 0 {
+			outputTokens = 0
+		}
+	}
```
irar2, line 396: Isn't it that if maxCompletionTokens is nil, this function should just return s.config.MaxModelLen?

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -375,6 +393,72 @@ func (s *VllmSimulator) isLora(model string) bool {
...
+
+// addRunningRequest adds a request to the running requests tracking
+func (s *VllmSimulator) addRunningRequest(reqID string, req completionRequest) {
+	processingTokens := s.calculateProcessingTokens(req)
+
+	runningReq := runningRequest{
+		promptTokens: req.getNumberOfPromptTokens(),
+		maxTokens:    processingTokens,
+		totalTokens:  processingTokens,
+	}
```
irar2: I could only find where 'totalTokens' is used (to update processingTokensCount), why 'promptTokens' and 'totalTokens'  are needed? And if they are not used, I guess we don't need runningRequestsMap at all? (And requestID)

irar2 (changes requested):

irar2:
@mohitpalsingh Thank you very much for this PR. After review of this PR along with issue #83 with a colleague of mine, we need to think about what exactly needs to be simulated here more carefully. Therefore, there will be a delay in the review of your PR.

pkg/llm-d-inference-sim/config_test.go
mohitpalsingh: done

pkg/llm-d-inference-sim/simulator.go
mohitpalsingh: resolved

pkg/llm-d-inference-sim/simulator.go
mohitpalsingh: You're right, that was unnecessary, I was biased on the general one fit for all use case approach in-case we would like further control over parallel requests in near future. But I guess removing it for now is better and leaner.
 
**Removed unused fields and structures:**
- runningRequest struct - Completely removed since promptTokens and maxTokens were never used
- runningRequestsMap sync.Map - Removed since we don't need to map request IDs to token counts
- requestID field - Removed from completionReqCtx since we no longer need unique request tracking

**Simplified token tracking:**
- Before: Store a complex runningRequest struct with 3 fields in a map, indexed by requestID
- After: Store just the processingTokens count directly in the completionReqCtx

**Updated method signatures:**
- addRunningRequest() now takes *completionReqCtx instead of (reqID, req)
- removeRunningRequest() now takes *completionReqCtx instead of reqID
- Both methods are simpler and more direct

PR #100, opened 2025-07-17
Enable configuration of various parameters in tools

### LINKED ISSUE #63: Enable configuration of various parameters
1. The simulator should support reading configuration settings from a file as well as inline parameters. Defining all parameters through the command line has become impractical due to their growing number and complexity. A file-based approach will make configuration more manageable, easier to update, and better suited for version control

2. We currently use hard coded values in many places in the code, e.g.,
- tool call arguments range
- tool call number of arguments
-  we "flip a coin" when we decide whether to add a not required parameter or field in tool call arguments
 We need to make these values configurable.

3. Consider supporting additional vLLM parameters, e.g., seed.

irar2:
Fix #63

### DIFF
--- a/pkg/llm-d-inference-sim/config.go
+++ b/pkg/llm-d-inference-sim/config.go
@@ -81,11 +81,11 @@ type configuration struct {
 	// optional, defaults to 1
 	MinToolCallArrayParamLength int `yaml:"min-tool-call-array-param-length"`
 
-	// ToolCallNotRequiredParamProbability is the probability to add a not required parameter in a tool call,
-	// optional, defaults to 50
+	// ToolCallNotRequiredParamProbability is the probability to add a parameter, that is not required,
+	// in a tool call, optional, defaults to 50
 	ToolCallNotRequiredParamProbability int `yaml:"tool-call-not-required-param-probability"`
-	// ObjectToolCallNotRequiredParamProbability is the probability to add a not required field in an object
-	// parameter in a tool call, optional, defaults to 50
+	// ObjectToolCallNotRequiredParamProbability is the probability to add a field, that is not required,
+	// in an object in a tool call, optional, defaults to 50
 	ObjectToolCallNotRequiredParamProbability int `yaml:"object-tool-call-not-required-field-probability"`
 }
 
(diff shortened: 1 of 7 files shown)


pkg/llm-d-inference-sim/config.go
```diff
@@ -174,5 +207,26 @@ func (c *configuration) validate() error {
...
+	if c.MaxToolCallIntegerParam < c.MinToolCallIntegerParam {
+		return errors.New("MaxToolCallIntegerParam cannot be less than MinToolCallIntegerParam")
+	}
+	if c.MaxToolCallNumberParam < c.MinToolCallNumberParam {
+		return errors.New("MaxToolCallNumberParam cannot be less than MinToolCallNumberParam")
+	}
+	if c.MaxToolCallArrayParamLength < c.MinToolCallArrayParamLength {
+		return errors.New("MaxToolCallArrayParamLength cannot be less than MinToolCallArrayParamLength")
+	}
+	if c.MaxToolCallArrayParamLength < 0 {
```
shmuelk: No need to check. If min is not less than zero, max can't be less than zero

pkg/llm-d-inference-sim/config_test.go
```diff
@@ -192,70 +192,103 @@ var _ = Describe("Simulator configuration", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(config).To(Equal(test.expectedConfig))
+			})
+		})
+	}
+
 	// Invalid configurations
+	invalidTests := make([]testCase, 0)
+
 	test = testCase{
```
shmuelk: The array of tests and invalid tests can be statically initialized

shmuelk (changes requested):

pkg/llm-d-inference-sim/config.go
irar2: Removed

pkg/llm-d-inference-sim/config_test.go
irar2: Done for invalid tests. For valid tests there is a call to newConfig(), so it is not static

shmuelk (approved):
/lgtm

/approve

PR #103, opened 2025-07-20
Choose latencies randomly

### LINKED ISSUE #101: Timing randomization using statistical parameters
To introduce more realistic timing behavior, all time-related parameters should support randomized values based on a mean and standard deviation, defined in the configuration. This approach will better simulate natural variability in processing times.

irar2:
Fix #101


dmitripikus (approved):
The changes are OK

PR #104, opened 2025-07-24
Support long responses and additional fixes

mayabar:
Currently, responses are generated by randomly selecting one sentence from a predefined list. As a result, even when max_token is set to a large value, long messages are never returned.

The suggestion to use a lorem ipsum generator was declined because the library we identified can produce only up to 191 words; any request for a larger word count causes it to panic.

New behavior:
- **Echo mode**: The input text is returned as-is, if `max_tokens` or `max_completion_tokens` is set to value higher than the input tokens number, if `max_tokens` or `max_completion_tokens` is lower - the input will be trimmed. Useful for testing where we need to know exact response in advance.
- **Random mode**:
  - if `max_tokens` or `max_completion_tokens` is specified, sentences are selected from the predefined collection until the token count reaches the limit
  - if number of output tokens is not defined in the request, a random token count is generated for the response. This count is based on a Gaussian distribution with mean of 40 and standard deviation of 20, capped by a maximum response length which is currently set to 128 tokens.

**Additional changes**: 
- Use tokenize function which divide text by space and additional characters in request processing too (not only in tools related part)
- Validate `max_token` and `max_completion_token` as request arrived and return 400 status if it's invalid
- Protect generating any random value with mutex
- Fix test for the changes above + add test for random texts creation

### DIFF
--- a/pkg/llm-d-inference-sim/utils.go
+++ b/pkg/llm-d-inference-sim/utils.go
@@ -101,7 +101,7 @@ func getRandomFinishReason() string {
 	return lengthFinishReason
 }
 
-// generateRandomText generates random text for the required number of tokens,
+// getRandomText generates random text for the required number of tokens,
 // select randomly a sentence from chatCompletionFakeResponses,
 // if number of tokens is lower than required - select another sentence,
 // continue until the required number of tokens is achieved


pkg/llm-d-inference-sim/utils_test.go
```diff
@@ -92,4 +115,17 @@ var _ = Describe("Utils", Ordered, func() {
 			Expect(totalTokens).Should(Equal(int64(250)))
 		})
 	})
+
+	Context("GetRandomText", func() {
+		var lenArr []int = []int{5, 20, 50, 150}
```
nekomeowww:
```suggestion
		lenArr := []int{5, 20, 50, 150}
```

pkg/llm-d-inference-sim/simulator_test.go
```diff
@@ -132,17 +139,15 @@ var _ = Describe("Simulator", func() {
...
 			Expect(numberOfChunksWithUsage).To(Equal(1))
-			Expect(chunk.Usage.PromptTokens).To(Equal(int64(4)))
+			Expect(chunk.Usage.PromptTokens).To(Equal(userMsgTokens))
 			Expect(chunk.Usage.CompletionTokens).To(BeNumerically(">", 0))
 			Expect(chunk.Usage.TotalTokens).To(Equal(chunk.Usage.PromptTokens + chunk.Usage.CompletionTokens))
 
 			msg := strings.Join(tokens, "")
-			expectedMsg := userMessage
-			if mode == modeRandom {
-				expectedMsg = getFullTextFromPartialString(msg)
```
shmuelk, line 142: Why were the comparisons of the randomly generated text removed? The comparison must now be done "sentence by sentence" in the generated code.

This comment  applies to many places in the code where getFullTextFromPartialString used to be invoked and was deleted.

shmuelk (changes requested):

pkg/llm-d-inference-sim/utils.go
```diff
@@ -72,29 +81,90 @@ func validateContextWindow(promptTokens int, maxCompletionTokens *int64, maxMode
...
+
+// getRandomFinishReason returns finish reason with the probability for 'stop' as defined by stopFinishReasonProbability
+func getRandomFinishReason() string {
+	if rand.Float64() < stopFinishReasonProbability {
+		return stopFinishReason
+	}
+	return lengthFinishReason
+}
+
+// generateRandomText generates random text for the required number of tokens,
```
nekomeowww:
```suggestion
// getRandomText generates random text for the required number of tokens,
```

mayabar: done

shmuelk (approved):
/lgtm

/approve

pkg/llm-d-inference-sim/simulator_test.go
mayabar: done

PR #105, opened 2025-07-29
Change project structure - separate main package to three

mayabar:
Separate llm-d-inference-sim to 3 packages: llmdinferencesim, common, openaiserverapi:

- common package contains general utilities and configuration class
- openaiserverapi package contains classes representing input and output structures for v1/chat/completions and v1/completions APIs with relevant utilities
- llmdinferencesim package contains the simulator startup and execution implementation

In addition rename config_test to simulator_stratup_test


pkg/common/utils.go
```diff
@@ -243,6 +248,44 @@ func init() {
...
+			}
+		}
+
+		if !found {
+			return false
+		}
+	}
+
+	return true
+}
```
shmuelk: The above code is test code and should not be in "non-test" file.

A better solution is to move the code to the file test/utils/utils.go

The appropriate import might be `testutils "github.com/llm-d/llm-d-inference-sim/test/utils"

shmuelk (changes requested):

pkg/common/utils.go
mayabar: Moving IsValidText to /test/utils creates a circular dependency since this function uses chatCompletionFakeResponses, which is defined in comon.
This function was moved to /pkg/comon/test-helpers.go which is used from tests only (utils_test and simulator_test), code that is not used in the non-test flow is not included to the executable.

shmuelk (approved):
/lgtm

/approve

PR #109, opened 2025-07-31
Code reorganization: moved configuration related code to common

### LINKED ISSUE #108: Code refactoring
Consider moving `parseCommandParamsAndLoadConfig` to the package `common`, along with the tests.

irar2:
Fixes #108

### DIFF
--- a/pkg/common/config.go
+++ b/pkg/common/config.go
@@ -151,7 +151,7 @@ func (c *Configuration) unmarshalLoras() error {
 	return nil
 }
 
-func NewConfig() *Configuration {
+func newConfig() *Configuration {
 	return &Configuration{
 		Port:                                vLLMDefaultPort,
 		MaxLoras:                            1,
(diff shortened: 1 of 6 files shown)


pkg/common/config.go
```diff
@@ -265,3 +268,110 @@ func (c *Configuration) Validate() error {
 	}
 	return nil
 }
+
+// ParseCommandParamsAndLoadConfig loads configuration, parses command line parameters, merges the values
+// (command line values overwrite the config file ones), and validates the configuration
+func ParseCommandParamsAndLoadConfig() (*Configuration, error) {
+	config := NewConfig()
```
mayabar: NewConfig could be a private function

mayabar (changes requested):

pkg/common/config.go
irar2: Done. It was used in simulator test, but I changed the test.

mayabar (approved):

PR #144, opened 2025-08-19
Support fake metrics

### LINKED ISSUE #138: Support for fake metrics
**Motivation**
For some e2e tests there is a requirement to create an environment with specific metrics in order to force specific use-cases 

**Implementation**
- define fake metrics values in configuration - support both command line and config file
- no partial fake metrics - never report real and fake metrics together
- metrics values that are not in the configuration are reported as their default values

irar2:
fixes #138

### DIFF
--- a/pkg/common/config.go
+++ b/pkg/common/config.go
@@ -34,6 +34,7 @@ const (
 	vLLMDefaultPort = 8000
 	ModeRandom      = "random"
 	ModeEcho        = "echo"
+	dummy           = "dummy"
 )
 
 type Configuration struct {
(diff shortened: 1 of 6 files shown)


pkg/common/config.go
```diff
@@ -127,6 +128,31 @@ type Configuration struct {
...
+
+	// FakeMetrics is a set of metrics to send to Prometheus instead of the real data
+	FakeMetrics *Metrics `yaml:"fake-metrics"`
+}
+
+type Metrics struct {
+	// LoraMetrics
+	LoraMetrics []LorasMetrics `json:"loras"`
+	LorasString []string       `yaml:"loras"`
+	// LorasJSONString []string `json:"loras"`
```
mayabar: commented out code ;)

irar2: fixed

PR #147, opened 2025-08-19
Use kv-cache-manager version v0.2.1

mayabar:
Fix serialization of BlockStored and BlockRemoved structures to be compatible to v0.2.1


shmuelk (approved):
/lgtm

/approve

Issue #141, opened 2025-08-19
Add retries to connect to ZMQ


mayabar:
Add max number of retries to the configuration

zhengkezhou1:
Hi @mayabar, I'm very interested in this project. Could you please assign this issue to me?

PR #152, opened 2025-08-20
Add ZMQ connection retry configuration

zhengkezhou1:
fix: #141

introduces a configurable number of retries for ZMQ connections. Previously, connection failures would terminate the process, but this new setting allows for more resilient network communication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ For more details see the <a href="https://docs.vllm.ai/en/stable/getting_started
 - `tokenizers-cache-dir`: the directory for caching tokenizers
 - `hash-seed`: seed for hash generation (if not set, is read from PYTHONHASHSEED environment variable)
 - `zmq-endpoint`: ZMQ address to publish events
-- `zmq-retries-times`: the maximum number of ZMQ connection attempts.
+- `zmq-max-connect-attempts`: the maximum number of ZMQ connection attempts. defaults to 0. maximum: 10
 - `event-batch-size`: the maximum number of kv-cache events to be sent together, defaults to 16
 - `fake-metrics`: represents a predefined set of metrics to be sent to Prometheus as a substitute for the actual data. When specified, only these fake metrics will be reported — real metrics and fake metrics will never be reported simultaneously. The set should include values for 
     - `running-requests`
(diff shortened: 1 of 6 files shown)


irar2:
@zhengkezhou1 Please also add a test that actually checks the retries:
Don't start the sub, and try to start the pub, then after a delay start the pub and check that the publisher is successfully created.

Also, please validate that the number of retries is not negative (it can be set to a negative value in a config file).
And add a test that checks that invalid values are rejected (in config_test.go) both from command line and from a yaml file.

pkg/common/config_test.go
```diff
@@ -377,6 +379,10 @@ var _ = Describe("Simulator configuration", func() {
 			args: []string{"cmd", "--fake-metrics", "{\"running-requests\":10,\"waiting-requests\":30,\"kv-cache-usage\":40}",
 				"--config", "../../manifests/config.yaml"},
 		},
+		{
+			name: "invalid ZMQ retries time",
+			args: []string{"cmd", "zmq-retries-times", "-1", "--config", "../../manifests/config.yaml"},
+		},
```
zhengkezhou1: I added a scenario here for when `retries` is a negative number.

README.md
```diff
@@ -122,6 +122,7 @@ For more details see the <a href="https://docs.vllm.ai/en/stable/getting_started
 - `tokenizers-cache-dir`: the directory for caching tokenizers
 - `hash-seed`: seed for hash generation (if not set, is read from PYTHONHASHSEED environment variable)
 - `zmq-endpoint`: ZMQ address to publish events
+- `zmq-retries-times`: the maximum number of ZMQ connection attempts.
```
irar2: Please rename to --zmq-max-connect-attempts and add a comment about default and max allowed values

pkg/common/config_test.go
irar2: Great, please also check what happens when it is set to a negative number in config file

README.md
zhengkezhou1: updated.

pkg/common/config_test.go
zhengkezhou1: i add a new file `invalid-config.yaml` to address this. also test default value(0) in `config.yaml`

PR #154, opened 2025-08-21
Show final config in simulaor default logger at Info lvel

### LINKED ISSUE #139: Print real merged configuration to log on simulator initialization

pancak3:
resolve #139

### DIFF
--- a/pkg/llm-d-inference-sim/simulator.go
+++ b/pkg/llm-d-inference-sim/simulator.go
@@ -20,6 +20,7 @@ package llmdinferencesim
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -717,8 +718,7 @@ func (s *VllmSimulator) getDisplayedModelName(reqModel string) string {
 
 func (s *VllmSimulator) showConfig(tgtLgr logr.Logger) error {
 	if tgtLgr == logr.Discard() {
-		err := fmt.Errorf("target logger is nil, cannot show configuration")
-		return err
+		return errors.New("target logger is nil, cannot show configuration")
 	}
 	cfgJSON, err := json.Marshal(s.config)
 	if err != nil {


irar2:
No need for a mock logger.

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -708,3 +714,17 @@ func (s *VllmSimulator) getDisplayedModelName(reqModel string) string {
...
+	if tgtLgr == logr.Discard() {
+		err := fmt.Errorf("target logger is nil, cannot show configuration")
+		return err
+	}
+	config := s.config
+	cfgJSON, err := json.MarshalIndent(config, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal configuration to JSON: %w", err)
+	}
+	tgtLgr.Info("Final simulator configuration:", "config", string(cfgJSON))
```
irar2: Please change tgtLgr.Info("Final simulator configuration:", "config", string(cfgJSON)) to tgtLgr.Info("Configuration:", "", string(cfgJSON))

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -708,3 +714,17 @@ func (s *VllmSimulator) getDisplayedModelName(reqModel string) string {
 	}
 	return s.config.ServedModelNames[0]
 }
+
+func (s *VllmSimulator) showConfig(tgtLgr logr.Logger) error {
+	if tgtLgr == logr.Discard() {
+		err := fmt.Errorf("target logger is nil, cannot show configuration")
+		return err
+	}
+	config := s.config
```
irar2: No need for a local var.

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -708,3 +714,17 @@ func (s *VllmSimulator) getDisplayedModelName(reqModel string) string {
...
+	cfgJSON, err := json.MarshalIndent(config, "", "  ")
+	if err != nil {
```
irar2, line 724: Two issues with this:
1. Most of the fields in the configuration only have yaml tags, and we see them as field names: 
"EnableKVCache": false, 
we want that to be consistent with what the user defines, so, please add json tags for Configuration fields:
EnableKVCache bool \`yaml:"enable-kvcache" json:"enable-kvcache"\`
2. There some fields that we use for unmarshalling, e.g., LoraModulesString, you can remove them by converting the json to a map, something like this:
`	var m map[string]interface{}
	err = json.Unmarshal(jsonBytes, &m)...
	delete(m, "LoraModulesString")`
	And also for LorasString inside FakeMetrics:
	`if fieldToRemove, ok := m["FakeMetrics"].(map[string]interface{}); ok {
		delete(fieldToRemove, "LorasString")
	}`
	And create a json to print:
	`cfgJSON, err := json.MarshalIndent(m, "", "  ")`

pancak3: For 2, I am thinking of creating a struct for log purposes. Then, perhaps convert the config to `ConfigurationLog` and marshal to indented JSON. What do you think?

Or is it better to separate the unmarshalling use fields? I anticipate they are `ServedModelNames` and `LoraModulesString`. If so, struct pairs `ConfigurationBase` within `Configuration` / `Configuration` within `ConfigurationParse` could be used. 

Related lines:
https://github.com/llm-d/llm-d-inference-sim/blob/4076bd2e5542d5f0a6fc0d4802a5cb98926b917d/pkg/common/config.go#L58
https://github.com/llm-d/llm-d-inference-sim/blob/4076bd2e5542d5f0a6fc0d4802a5cb98926b917d/pkg/common/config.go#L46

pancak3: For 1, IIUC, this issue(#139) aims to show only the final configuration. @mayabar Maya, is there any preferred format of the log? yaml, perhaps, so user can copy and paste.

The current  output is json:
<details>
<summary>(click to see)</summary>

```
I0821 11:31:34.836881   44692 main.go:36] "Starting vLLM simulator"
I0821 11:31:34.840546   44692 simulator.go:727] "Configuration:" =<
	{
	  "port": 8000,
	  "model": "repo/model",
	  "served-model-name": [
	    "repo/model"
	  ],
	  "max-loras": 1,
	  "max-cpu-loras": 1,
	  "max-num-seqs": 5,
	  "max-model-len": 1024,
	  "lora-modules": [
	    "{\"name\":\"tweet-summary-0\"}",
	    "{\"name\":\"tweet-summary-1\"}"
	  ],
	  "LoraModules": [
	    {
	      "name": "tweet-summary-0",
	      "path": "",
	      "base_model_name": ""
	    },
	    {
	      "name": "tweet-summary-1",
	      "path": "",
	      "base_model_name": ""
	    }
	  ],
	  "time-to-first-token": 0,
	  "time-to-first-token-std-dev": 0,
	  "inter-token-latency": 0,
	  "inter-token-latency-std-dev": 0,
	  "kv-cache-transfer-latency": 0,
	  "kv-cache-transfer-latency-std-dev": 0,
	  "mode": "random",
	  "seed": 1755775894840429416,
	  "max-tool-call-integer-param": 100,
	  "min-tool-call-integer-param": 0,
	  "max-tool-call-number-param": 100,
	  "min-tool-call-number-param": 0,
	  "max-tool-call-array-param-length": 5,
	  "min-tool-call-array-param-length": 1,
	  "tool-call-not-required-param-probability": 50,
	  "object-tool-call-not-required-field-probability": 50,
	  "enable-kvcache": false,
	  "kv-cache-size": 1024,
	  "tokenizers-cache-dir": "",
	  "block-size": 16,
	  "hash-seed": "",
	  "zmq-endpoint": "tcp://localhost:5557",
	  "event-batch-size": 16,
	  "fake-metrics": null
	}
 >
```
</details>

What more we can do is to show both argument style`--xxx yyy` and yaml. or the one which was inputed.

mayabar: > For 2, I am thinking of creating a struct for log purposes. Then, perhaps convert the config to `ConfigurationLog` and marshal to indented JSON. What do you think?

I think that defining an additional structure is redundant work, in addition, any new configuration field should be added to both places, which can lead to inconsistency if someone misses the second place.

> Or is it better to separate the unmarshalling use fields? I anticipate they are `ServedModelNames` and `LoraModulesString`. If so, struct pairs `ConfigurationBase` within `Configuration` / `Configuration` within `ConfigurationParse` could be used.

`ServedModelNames` is a valid field that should be logged with other fields.
Special fields that should be removed are `LoraModulesString` and `LorasString` under `fake-metrics`.
I would not change structures specially for logging prepuces. Please add a function that embeds the processing of the original config and printing it to the log, like Ira mentioned.

mayabar: > For 1, IIUC, this issue(#139) aims to show only the final configuration. @mayabar Maya, is there any preferred format of the log? yaml, perhaps, so user can copy and paste.

I think the output you have is great. Presenting the config as a json with field names exactly as in the command line parameters/yaml file field name will be easy for the user to understand.
Regarding the example you presented above:
- "lora-modules": [
	    "{\"name\":\"tweet-summary-0\"}",
	    "{\"name\":\"tweet-summary-1\"}"
	  ] should be removed
- "LoraModules": should be "lora-modules": [
To be sure that `LorasString ` is removed too, please define "fake-metrics" object in config file or command line parameters.

> What more we can do is to show both argument style`--xxx yyy` and yaml. or the one which was inputed.

I think that presenting the final value is valuable enough, no need to write from which source (default/config file/command line param) the value is taken

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -708,3 +714,53 @@ func (s *VllmSimulator) getDisplayedModelName(reqModel string) string {
...
+		if err != nil {
+			return fmt.Errorf("failed to marshal fake-metrics to JSON: %w", err)
+		}
+		err = json.Unmarshal(fakeMetricsJSON, &fakeMetricsM)
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal fake-metrics to map: %w", err)
+		}
+		delete(fakeMetricsM, "LorasString")
+		// set fake-metrics
+		m["fake-metrics"] = fakeMetricsM
```
irar2: Why did you use marshal and unmarshal here, why not use 
`if fieldToRemove, ok := m["FakeMetrics"].(map[string]interface{}); ok { delete(fieldToRemove, "LorasString") }`?

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -708,3 +714,53 @@ func (s *VllmSimulator) getDisplayedModelName(reqModel string) string {
...
+	var m map[string]interface{}
+	err = json.Unmarshal(cfgJSON, &m)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal JSON to map: %w", err)
+	}
+	m["lora-modules"] = m["LoraModules"]
+	delete(m, "LoraModules")
+	if m["lora-modules"] == nil {
+		m["lora-modules"] = ""
+	}
```
irar2: If an array is empty, nil is ok

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -708,3 +714,53 @@ func (s *VllmSimulator) getDisplayedModelName(reqModel string) string {
...
+		err = json.Unmarshal(fakeMetricsJSON, &fakeMetricsM)
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal fake-metrics to map: %w", err)
+		}
+		delete(fakeMetricsM, "LorasString")
+		// set fake-metrics
+		m["fake-metrics"] = fakeMetricsM
+	} else {
+		m["fake-metrics"] = ""
+	}
```
irar2, line 741: If an array is empty, nil is ok

irar2 (approved):

PR #169, opened 2025-08-25
feat: generate response length based on a histogram when max_tokens is defined in the request

### LINKED ISSUE #167: Enhance calculation of tokens number in response based on request's max_tokens parameter - simple logic

mayabar:
- Added logic to randomly select the response length based on a pre-defined histogram when the request includes the max_tokens property
- This initial version applies the same histogram regardless of the specific max_tokens value
- Future improvements will introduce dynamic buckets adapting the histogram according to different max_tokens values.

fix #167

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -39,7 +39,7 @@ const (
 	RemoteDecodeFinishReason = "remote_decode"
 )
 
-// this array defines probabilities for buckets used for generation of number of tokens in response
+// this array defines the probabilities for the buckets to be used for the generation of number of tokens in response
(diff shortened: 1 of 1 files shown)


pkg/common/utils.go
```diff
@@ -154,14 +167,62 @@ func GetRandomResponseText(maxCompletionTokens *int64) (string, string) {
...
+
+	r := RandomFloat(0, 1)
+
+	// probability to return maxToken
+	if r > cumulativeBuckets[len(cumulativeBuckets)-2] {
+		return maxTokens
+	}
+
+	// determine which bucket to use
+	bucketIndex := 0
```
shmuelk: Shouldn't this be `bucketIndex := len(cumulativeBuckets)-1` ?

PR #170, opened 2025-08-25
Use dynamic ports in zmq tests

### LINKED ISSUE #164: ZMQ sub *:port failed randomly in CI
**What happened**:
KV cache test failed *sometimes*;  ZQM tries to sub but the address is in use. It happens in CI.

See jobs:
https://github.com/llm-d/llm-d-inference-sim/actions/runs/17185865922/job/48754776272#step:8:131
https://github.com/llm-d/llm-d-inference-sim/actions/runs/17186323665/job/48755883796

Line:
https://github.com/llm-d/llm-d-inference-sim/blob/bdc5ecb6e6b331f99000640e14aa0f509e18f44b/pkg/kv-cache/kv_cache_test.go#L39

Guess:
The executor that runs the job has side process occupied the addr `*:5557`

Proposal:
@pancak3 - use `127.0.0.1` in tests for both sub and pub
@irar2  - look for available port

**What you expected to happen**:
Pass, successfully sub the addr

**How to reproduce it (as minimally and precisely as possible)**:
Trigger ci go test, it *may* fail

**Anything else we need to know?**:
N/A

**Environment**:
N/A

pancak3:
resolve #164

### DIFF
--- a/pkg/common/publisher_test.go
+++ b/pkg/common/publisher_test.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"encoding/binary"
 
-	"net"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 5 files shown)


irar2:
Looking at https://libzmq.readthedocs.io/en/zeromq4-1/zmq_tcp.html, it looks like the address to start should be "tcp://*:*" (please make it a constant)

pancak3:
> Looking at https://libzmq.readthedocs.io/en/zeromq4-1/zmq_tcp.html, it looks like the address to start should be `tcp://*:*` (please make it a constant)

How about `tcp://127.0.0.1:*`? Listening on all interface sounds too open for test purpose

irar2:
The tests fail in https://github.com/pancak3/llm-d-inference-sim/blob/fab7019849fdcdfba989d4f66b3ecd230b53e9d6/pkg/kv-cache/kv_cache_test.go#L421
because there is no sub there, the solution for this is to prevent publishing events by not updating ZMQEndpoint in the configuration, and in this case there is no need to create a publisher (in newBlockCache) and then in KVEventSender.Run we should continue if there is no publisher.

pkg/kv-cache/kv_cache_sender.go
```diff
@@ -159,6 +159,11 @@ func (s *KVEventSender) publishHelper(ctx context.Context) error {
 		DataParallelRank: &dpRank,
 	}
 
+	if s.publisher == nil {
```
irar2: This check should be in Run() in two cases: 
`case eventData, ok := <-s.eventChan:` 
after `if !ok{}` please add in line 90
`if s.publisher == nil {
				continue
			}
`
and
`case <-timer.C:` 
please add the same check first thing in this case.

I don't think we need a log message here, there can be too many.

pkg/common/publisher_test.go
```diff
@@ -42,7 +41,9 @@ var _ = Describe("Publisher", func() {
 		Expect(err).NotTo(HaveOccurred())
 		sub, err := zctx.NewSocket(zmq.SUB)
 		Expect(err).NotTo(HaveOccurred())
-		err = sub.Bind(subEndpoint)
+		err = sub.Bind(endpoint)
+		Expect(err).NotTo(HaveOccurred())
+		endpoint, err := sub.GetLastEndpoint()
```
irar2, line 46: Could you please use different names for endpoint, maybe call the initial one `wildcardEndpoint`? It's a bit confusing.

irar2 (approved):

irar2:
@pancak3 Thanks!

/lgtm
/approve

PR #177, opened 2025-09-02
Bugfix: was accessing number of tokens from nil var; getting it from req instead

pancak3:


pancak3:
@mayabar @irar2 Apologies for introducing this bug. This is an urgent fix.

PR #184, opened 2025-09-03
Take cached prompt tokens into account in prefill time calculation

### LINKED ISSUE #178: KV Cache simulation for calculation of request prefill time
For a request, check if we already have a part of it in the local kv cache, and only use the remaining number of tokens in prefill time calculation

irar2:
Closes #178 

In addition, return error if tokenization fails.
Small code refactoring.

### DIFF
--- a/pkg/llm-d-inference-sim/simulator_test.go
+++ b/pkg/llm-d-inference-sim/simulator_test.go
@@ -879,7 +879,7 @@ var _ = Describe("Simulator", func() {
 			Entry("large overhead, 1024 tokens", 2000, 3000, 800, 1024, 0),
 			Entry("very long prompt", 150, 200, 70, 20000, 0),
 			Entry("medium overhead, 512 tokens, 256 cached", 200, 1000, 150, 512, 256),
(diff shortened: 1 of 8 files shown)


pkg/llm-d-inference-sim/simulator_test.go
```diff
@@ -851,33 +851,60 @@ var _ = Describe("Simulator", func() {
...
+					prefillOverhead, prefillTimePerToken, stdDev, nTokens, nCachedTokens)
+			},
+			Entry("single token", 100, 50, 10, 1, 0),
+			Entry("single token big std", 100, 50, 70, 1, 0),
+			Entry("stddev is 0", 100, 50, 0, 1, 0),
+			Entry("medium overhead, 512 tokens", 200, 1000, 150, 512, 0),
+			Entry("large overhead, 1024 tokens", 2000, 3000, 800, 1024, 0),
+			Entry("very long prompt", 150, 200, 70, 20000, 0),
+			Entry("medium overhead, 512 tokens, 256 cached", 200, 1000, 150, 512, 256),
+			Entry("large overhead, 1024 tokens, 128 cached", 2000, 3000, 800, 1024, 1008),
```
mayabar: typo in name of the test

PR #187, opened 2025-09-04
Add ignore eos in request

### LINKED ISSUE #186: Respect ignore_eos in sim
When generate tokens, respect `ignore_eos`

**Why is this needed**:
https://github.com/llm-d/llm-d-inference-sim/blob/639b40eefcfaadb18ffe5a4c634a220d77c19115/pkg/common/utils.go#L176
Currently, when generating tokens, `max_tokens` or `max_completion_tokens` is used to validate the length of the generated output. However, the `ignore_eos` parameter is not taken into account.

When evaluating model performance on production request datasets, it's crucial to force models to generate a precise number of tokens for a fair comparison. Setting `ignore_eos` to **true** ensures that the output will be exactly `max_tokens` (or `max_completion_tokens`) long, mimicking the behavior of real-world inference services more accurately.

Therefore, we need to modify the token generation logic to properly handle the `ignore_eos` parameter. This will allow for consistent and reproducible evaluations by ensuring the number of output tokens is equal to the specified maximum length when `ignore_eos` is true.

pancak3:
/resolve #186

### DIFF
--- a/pkg/openai-server-api/request.go
+++ b/pkg/openai-server-api/request.go
@@ -59,6 +59,8 @@ type CompletionRequest interface {
 	GetToolChoice() string
 	// GetMaxCompletionTokens returns the maximum completion tokens requested
 	GetMaxCompletionTokens() *int64
+	// GetIgnoreEOS returns true if the end of sequence will be ignored
+	GetIgnoreEOS() bool
 	// IsDoRemoteDecode() returns true if do_remote_decode field is true in the request,
 	// when the field is true, the decode phase should be done on remote pod,
 	// whereas prefill phase is done on local pod, thus this is a prefill request
@@ -164,6 +166,9 @@ type ChatCompletionRequest struct {
 	// tokens and reasoning tokens.
 	MaxCompletionTokens *int64 `json:"max_completion_tokens"`
 
+	// IgnoreEOS is a boolean value, true when the model should ignore end-of-sequence tokens
+	IgnoreEOS bool `json:"ignore_eos"` // Field remains unchanged
+
 	// Tools is a list of tools the model may call.
 	Tools []Tool `json:"tools,omitempty"`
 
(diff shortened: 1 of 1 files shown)


pkg/common/utils.go
```diff
@@ -165,7 +165,7 @@ func GetRandomText(numOfTokens int) string {
 // if maxCompletionTokens is nil
 // - the response text's length is randomly chosen from the range [1, responseLenMax] according additional parameters
 // - finish reason is stop
-func GetRandomResponseText(maxCompletionTokens *int64) (string, string) {
+func GetRandomResponseText(maxCompletionTokens *int64, ignore_eos bool) (string, string) {
```
mayabar, line 170: please add explanation of the new parameter to the comment

pkg/openai-server-api/request.go
```diff
@@ -164,6 +166,9 @@ type ChatCompletionRequest struct {
 	// tokens and reasoning tokens.
 	MaxCompletionTokens *int64 `json:"max_completion_tokens"`
 
+	// IgnoreEOS is a boolean value, true when the model should ignore end-of-sequence tokens
+	IgnoreEOS bool `json:"ignore_eos"` // Field remains unchanged
```
mayabar: what do you mean by "Field remains unchanged"?

pkg/openai-server-api/request.go
```diff
@@ -264,6 +273,9 @@ type TextCompletionRequest struct {
 	// The token count of your prompt plus `max_tokens` cannot exceed the model's
 	// context length.
 	MaxTokens *int64 `json:"max_tokens"`
+
+	// IgnoreEOS is a boolean value, true when the model should ignore end-of-sequence tokens
+	IgnoreEOS bool `json:"ignore_eos"`
```
mayabar: IgnoreEOS is defined in both TextCompletionRequest and ChatCompletionRequest, please move to baseCompletionRequest

mayabar (changes requested):

PR #189, opened 2025-09-04
Validate max-num-seqs

### LINKED ISSUE #157: Adjust request "processing time" to current load
**Current state**:
Request processing time is composed of the prefill and decode parts. 
If prefill is executed on the same instance, its time can be defined by one of the following options:
- `time-to-first-token (TTFT)` parameter: represents prefill time independent of prompt size
- `prefill-overhead` + `prefill-time-per-token (PTPT)` * `number-of-propmpt-tokens`: prefill time which takes into account the prompt length
In case of disaggregated prefill (prefill executed on a different pod), the time is calculated as `kv-cache-transfer-latency` * `number-of-propmpt-tokens`

Decode time is calculated as:
`inter-token-latency` * `number-of-propmpt-tokens`

Allow support to adjust all the above timeouts based on the current load (i.e., the number of requests currently running in parallel)
Introduce a new configuration parameter `load-factor` which defines the multiplier for all time delays when the load is at maximum. 
All time configuration values should be calculated using linear interpolation between the number of requests [1-max-num-seq] and the time range [time-defined-in-config, time-defined-in-config*load-factor]

The formula:
time = time_defined_in_config * [1 + (load_factor - 1) * (current_requests_num - 1) / (max_num_seq - 1)]

**Why is this needed**:
The real vLLM's performance depends on the current load. The simulator should mimic this behavior for realistic results.

pancak3:
/resolve #157

### DIFF
--- a/pkg/common/config.go
+++ b/pkg/common/config.go
@@ -373,6 +373,10 @@ func (c *Configuration) validate() error {
 		return errors.New("max model len cannot be less than 1")
 	}
 
+	if c.MaxNumSeqs < 1 {
+		return errors.New("max num seqs cannot be less than 1")
+	}
+
 	for _, lora := range c.LoraModules {
 		if lora.Name == "" {
 			return errors.New("empty LoRA name")
--- a/pkg/common/config_test.go
+++ b/pkg/common/config_test.go
@@ -431,6 +431,16 @@ var _ = Describe("Simulator configuration", func() {
 			args: []string{"cmd", "--data-parallel-size", "15",
 				"--config", "../../manifests/config.yaml"},
 		},
+		{
+			name: "invalid max-num-seqs",
+			args: []string{"cmd", "--max-num-seqs", "0",
+				"--config", "../../manifests/config.yaml"},
+		},
+		{
+			name: "invalid max-num-seqs",
+			args: []string{"cmd", "--max-num-seqs", "-1",
+				"--config", "../../manifests/config.yaml"},
+		},
 	}
 
 	for _, test := range invalidTests {


pkg/llm-d-inference-sim/simulator.go
```diff
@@ -708,16 +708,16 @@ func (s *VllmSimulator) getTimeToFirstToken(nPromptTokens int, nCachedPromptToke
 	}
 	if s.config.TimeToFirstToken == 0 && s.config.TimeToFirstTokenStdDev == 0 {
 		// is aggregated PD and ttft is calculated using number of prompt tokens that are not in kv cache
-		prefillTime := s.config.PrefillOverhead + (nPromptTokens-nCachedPromptTokens)*s.config.PrefillTimePerToken
+		prefillTime := s.config.GetPrefillOverhead(s.nRunningReqs) + (nPromptTokens-nCachedPromptTokens)*s.config.GetPrefillTimePerToken(s.nRunningReqs)
```
pancak3: @mayabar, is it better to move the methods from `config` to `simulator`? It makes sense since the results relies on the number of running requests, and `config` also belongs to `simulator`. This should simplify the code.

pkg/common/config.go
```diff
@@ -168,6 +176,29 @@ type Configuration struct {
...
+}
+
+func (c *Configuration) GetPrefillTimePerToken(nRunningReqs int64) int {
+	return int(float64(c.PrefillTimePerToken) * c.calcLoadFactor(nRunningReqs))
+}
+
+func (c *Configuration) GetInterTokenLatency(nRunningReqs int64) int {
+	return int(float64(c.InterTokenLatency) * c.calcLoadFactor(nRunningReqs))
+}
+
```
mayabar: we defined configuration as a data storage without logic related to the simulator current state, can you please move these functions to the simulator or utilities?

pkg/llm-d-inference-sim/simulator.go
mayabar: please see the comment above, we try to keep the configuration not mixed with simulator behavior

mayabar (changes requested):

PR #197, opened 2025-09-09
Wrap time sleep

pancak3:


mayabar:
Hi Qifan, @pancak3 
We don’t think we need to wrap `time.Sleep`. The direct call is clear enough and avoids introducing an extra layer of indirection.

PR #200, opened 2025-09-13
Use custom dataset as response source

### LINKED ISSUE #196: Support custom datasets
Allow user provided custom datasets for response generation. 

At the moment, the simulator has a short set of sentences from which it randomly chooses a response text. Support an option for user-provided custom dataset instead.

pancak3:
/resolve #196


pancak3:
Peview with [LibreChat](https://github.com/danny-avila/LibreChat) with [8c94372](https://github.com/llm-d/llm-d-inference-sim/pull/200/commits/8c943721b16847332e1b65186813f3d37b287c5a) as backend:

output

@mayabar @irar2 this pr is big. I refactored a lot while implementing the custom dataset. Please feel no hesitation to let me know anywhere to fix or improve

mayabar:
Hi Qifan, 
We briefly went over this PR and have some concerns about the custom dataset implementation. To make sure we understood the concept correctly, could you explain in a couple of sentences the main idea behind how a generated text is selected.
Additionally, do you have some utility for converting the HF dataset into the format you are using? And why not simply store it in a text file and load it into a map in a memory?

pancak3:
> Hi Qifan, We briefly went over this PR and have some concerns about the custom dataset implementation. To make sure we understood the concept correctly, could you explain in a couple of sentences the main idea behind how a generated text is selected. Additionally, do you have some utility for converting the HF dataset into the format you are using? And why not simply store it in a text file and load it into a map in a memory?

Hi Maya,

Yes I expected concerns. Perhaps I should have discussed about this before the implementation.

For a request, a hash is generated, full messages for chat and whatever prompt for text completion. The hash is used to lookup responses. If no response found, or number of tokens is greater than max tokens, the desired number of tokens will be used to lookup responses. The “desired number of tokens” is generated using existing code.

If there are responses found in dataset, randomly select one from them. Otherwise, randomly generate using existing implementation.

> And why not simply store it in a text file and load it into a map in a memory?

The main purpose of using a prepared dataset is for performance, with some other considerations.

1. Reduce load time and avoid repeating of computation. Using hash to lookup responses requires all the conversations have hashes. If generating them when starting llmd-sim, it takes linear time. The computation will also be repeated every time starting on every instances. This seems not eco for me.
2. Considerations for memory capacity. Loading all dataset to consume memory of the host machine, thus may make llmd-sim heavy. Relatively, disk is easier to obtain. Especially for users who work with giant datasets. But this is a trade off; memory is faster.
3. Others, decoupling the functionality for the datasets preparation. There are different datasets when they have very different structures. Splitting the converting in another place seems better for repo management, to me.

>  do you have some utility for converting the HF dataset into the format you are using? 

Yes, the utility is with the dataset in the huggingface repo. There is readme as well. 


---
 To wrap up, the implementation is mainly for performance. I can add a config to let user decide the behaviour, with “loading dataset” in memory. Let me know what you think.

README.md
```diff
@@ -149,7 +149,10 @@ For more details see the <a href="https://docs.vllm.ai/en/stable/getting_started
...
 ---
 - `data-parallel-size`: number of ranks to run in Data Parallel deployment, from 1 to 8, default is 1. The ports will be assigned as follows: rank 0 will run on the configured `port`, rank 1 on `port`+1, etc.      
-
+---
+- `dataset-path`: local path to the sqlite db file for response generation from a dataset, optional, if not set, but `dataset-url` is set, the default path `<USER_HOME>/.llm-d/dataset.sqlite3` will be used. If neither dataset-path nor dataset-url are set, response is randomly generated. See [llm-d converted ShareGPT](https://huggingface.co/datasets/llm-d/inference-sim-datasets/blob/0b7ac1a4daf0aace1556326964bd75633372299e/README.md) for more details on the expected format of the sqlite db file.
+- `dataset-url`: URL to download the sqlite db file for response generation from a dataset, optional, if set, the sqlite db file will be downloaded to the path specified by `dataset-path`. If the file already exists at that path, it will not be downloaded again. Example url: `https://huggingface.co/datasets/llm-d/inference-sim-datasets/resolve/fcffe19126878d9f3264fded2a0a44c4f16636b9/huggingface/ShareGPT_Vicuna_unfiltered/conversations.sqlite3`
```
mayabar: Please update links to HF

pkg/common/config.go
```diff
@@ -174,6 +174,19 @@ type Configuration struct {
 
 	// DPSize is data parallel size - a number of ranks to run, minimum is 1, maximum is 8, default is 1
 	DPSize int `yaml:"data-parallel-size" json:"data-parallel-size"`
+
+	// Dataset configuration for response generation from a dataset. sqlite db file is expected.
+	Dataset DatasetConf
```
mayabar: We keep all parameter in a flat list, can you please continue with this convension

pkg/common/config.go
```diff
@@ -543,6 +556,9 @@ func ParseCommandParamsAndLoadConfig() (*Configuration, error) {
 	f.IntVar(&config.EventBatchSize, "event-batch-size", config.EventBatchSize, "Maximum number of kv-cache events to be sent together")
 	f.IntVar(&config.DPSize, "data-parallel-size", config.DPSize, "Number of ranks to run")
 
+	f.StringVar(&config.Dataset.Path, "dataset-path", config.Dataset.Path, "Local path to the sqlite db file for response generation from a dataset")
```
mayabar: please keep consistency between command line parameter names and yaml parameters names, e.g. dataset-path vs path

pkg/dataset/custom_dataset.go
```diff
@@ -0,0 +1,418 @@
...
+	genTokensCol      = "gen_tokens"
+	nGenTokensCol     = "n_gen_tokens"
+	promptHashColType = "BLOB"
+	genTokensColType  = "JSON"
+	nGenTokensColType = "INTEGER"
+)
+
+func (d *CustomDataset) downloadDataset(url string, path string) error {
+	// Set up signal handling for Ctrl+C (SIGINT)
+	ctx, cancel := context.WithCancel(context.Background())
```
mayabar: there is a context created in the main, please use it

README.md
```diff
@@ -149,7 +149,10 @@ For more details see the <a href="https://docs.vllm.ai/en/stable/getting_started
       {"running-requests":10,"waiting-requests":30,"kv-cache-usage":0.4,"loras":[{"running":"lora4,lora2","waiting":"lora3","timestamp":1257894567},{"running":"lora4,lora3","waiting":"","timestamp":1257894569}]}
 ---
 - `data-parallel-size`: number of ranks to run in Data Parallel deployment, from 1 to 8, default is 1. The ports will be assigned as follows: rank 0 will run on the configured `port`, rank 1 on `port`+1, etc.      
-
+---
+- `dataset-path`: local path to the sqlite db file for response generation from a dataset, optional, if not set, but `dataset-url` is set, the default path `<USER_HOME>/.llm-d/dataset.sqlite3` will be used. If neither dataset-path nor dataset-url are set, response is randomly generated. See [llm-d converted ShareGPT](https://huggingface.co/datasets/llm-d/inference-sim-datasets/blob/0b7ac1a4daf0aace1556326964bd75633372299e/README.md) for more details on the expected format of the sqlite db file.
```
mayabar: consider to not use default value `<USER_HOME>/.llm-d/dataset.sqlite3`, in case of multiple runs with default value, only first time the dataset will be downloaded and all other executions will use the first dataset which is not necessary the right one (in `dataset-url` help written that is file already exists, the dataset will not be downloaded but simulator will run)

pkg/dataset/custom_dataset.go
```diff
@@ -0,0 +1,418 @@
...
+	defer signal.Stop(sigs)
+
+	// Goroutine to listen for signal
+	go func() {
+		<-sigs
+		d.Logger.Info("Interrupt signal received, cancelling download...")
+		cancel()
+	}()
+
+	out, err := os.Create(path)
```
mayabar, line 72: In the readme written that "If the file already exists at that path, it will not be downloaded again."
Here if the file already exists - seems that it will be opened, truncated to zero length and downloaded content will be written to it

pkg/dataset/dataset.go
```diff
@@ -0,0 +1,332 @@
...
+
+	if *maxCompletionTokens >= int64(len(tokens)) {
+		return tokens, StopFinishReason
+	}
+	// return truncated text
+	return tokens[0:*maxCompletionTokens], LengthFinishReason
+}
+
+type BaseDataset struct {
+	Logger logr.Logger
```
mayabar: logger can be a private member

pkg/dataset/custom_dataset.go
```diff
@@ -0,0 +1,418 @@
...
+		if cerr != nil {
+			d.Logger.Error(cerr, "failed to remove incomplete file after download")
+		}
+		// If context was cancelled, return a specific error
+		if errors.Is(err, context.Canceled) {
+			return errors.New("download cancelled by user")
+		}
+		return fmt.Errorf("failed to download file: %w", err)
+	}
+	// Check if file size is zero or suspiciously small
```
mayabar: "or suspiciously small" can be removed, we don't check this

pkg/dataset/custom_dataset.go
```diff
@@ -0,0 +1,418 @@
...
+	select {
+	case <-pr.ctx.Done():
+		return 0, pr.ctx.Err()
+	default:
+	}
+	n, err := pr.Reader.Read(p)
+	pr.downloaded += int64(n)
+	if pr.total > 0 {
+		pct := int(float64(pr.downloaded) * 100 / float64(pr.total))
+		if !pr.hasShownSpeed && time.Since(pr.startTime).Seconds() > 2 {
```
mayabar: hasShownSpeed allows to print the process statistics after 2 sec, consider another name that reflects the purpose

pkg/dataset/custom_dataset.go
```diff
@@ -0,0 +1,418 @@
...
+			pr.hasShownSpeed = true
+			pr.logProgress(pct)
+			pr.lastPct = pct
+		}
+		if pct != pr.lastPct && pct%10 == 0 {
```
mayabar: seems that it's possible that percentage will jump over number which is a multiplication of 10, e.g. from 19 to 21

mayabar (changes requested):

mayabar:
@pancak3 I just noticed that I forgot to finish review and all my comments were in pending state for last 3 days 🤦‍♀️

pancak3:
> @pancak3 I just noticed that I forgot to finish review and all my comments were in pending state for last 3 days 🤦‍♀️

It happens. :) This PR is big, so let me know when you've finished the review, and then I'll start to address them. Thanks!

mayabar:
I had pending comments, submitted them about 1 hour ago

README.md
mayabar: @pancak3 - today's comments start here :)

pancak3:
Hi Maya, I have resolved the above comments. Also added a config `dataset-in-memory` to switch whether to load the dataset in memory or not. Please review, thanks!

pkg/dataset/custom_dataset.go
```diff
@@ -246,11 +247,69 @@ func (d *CustomDataset) getRecordsCount() (int, error) {
 	return count, nil
 }
 
-func (d *CustomDataset) connectToDB(path string) error {
+func (d *CustomDataset) loadDatabaseInMemory(path string) error {
+	d.logger.Info("Loading database into memory...")
```
mayabar, line 251: maybe add here path of the loaded file?

pkg/dataset/custom_dataset_test.go
```diff
@@ -173,7 +176,7 @@ var _ = Describe("CustomDataset", Ordered, func() {
 	})
 
 	It("should return tokens for existing prompt", func() {
-		err := dataset.Init(validDBPath, "")
+		err := dataset.Init(context.Background(), logr.Discard(), validDBPath, "", false)
```
mayabar: usually we use logger := klog.Background() (see simulator_test for reference)

mayabar (commented):
@pancak3 thanks for your updates, I added some comments

pkg/dataset/custom_dataset.go
pancak3: It may duplicate later log

Log when in memory is used
```
I0929 10:47:49.827213   18376 custom_dataset.go:371] "Using dataset from" path=".llm-d/dataset.sqlite3"
I0929 10:47:49.827384   18376 custom_dataset.go:251] "Loading database into memory..."
I0929 10:48:00.800116   18376 custom_dataset.go:304] "Database loaded into memory" load_time="10.971650214s"
I0929 10:48:01.304345   18376 custom_dataset.go:357] "In-memory database connected successfully" path=".llm-d/dataset.sqlite3" records count=261914
I0929 10:48:01.306202   18376 server.go:37] "Server starting" port=8000
I0929 10:48:01.343105   18376 server.go:81] "Server starting" protocol="HTTP" port=8000
```
not used
```
I0929 10:49:36.923809   18801 custom_dataset.go:371] "Using dataset from" path=".llm-d/dataset.sqlite3"
I0929 10:49:37.480871   18801 custom_dataset.go:359] "Database connected successfully" path=".llm-d/dataset.sqlite3" records count=261914
I0929 10:49:37.481190   18801 server.go:37] "Server starting" port=8000
I0929 10:49:37.510321   18801 server.go:81] "Server starting" protocol="HTTP" port=8000
```

The `Server starting` log seems to be duplicated. I will include a commit in this pr to fix it.

mayabar: cool!

mayabar (approved):

PR #214, opened 2025-10-16
New requests queue

### LINKED ISSUE #172: Change requests queue implementation from a channel to more complex queue
Currently, requests queue is implemented using a go channel. This causes some limitations:

**Known limitation 1**: max-loras is not taken into consideration when worker pulls request from the queue
The component that receives requests, push a new request to a channel, workers are waiting for a request on this channel and process them.
Always requests are processed by arrival time.
In some scenarios current behavior leads to running requests in parallel with more lora adapters than defined in the max-lora parameter.

*Example:*
max-loras is deined to 2, number of parallel requests - 3
Queue contains: R1(lora1), R2(lora2), R3(lora3), R4 (lora1)
Current implementation will pull 3 requests for processing r1, r2, and r3.
Required behavior: pull r1, r2, r4 (r3 could not be sent for processing since it will cause loading of more than 2 loras)

**Known limitation 2**: extra entries in loraInfo metrics are reported
When a single request is received, it is pushed to the queue (the channel) which creates metrics report that the lora of this request is in waiting list. This report does not affects the LoraAwareScorer but should be removed in the future version. Implementing new queue will fix this behavior.

**Solution:**
Implement a queue class which will expose API similar to channel - workers will wait for a new request to process. 
It will skip requests with loras that cannot be loaded right now.
Design - **TBD**

irar2:
Closes #172

### DIFF
--- a/pkg/llm-d-inference-sim/simulator.go
+++ b/pkg/llm-d-inference-sim/simulator.go
@@ -116,8 +116,6 @@ type metricsData struct {
 	requestParamsMaxTokens *prometheus.HistogramVec
 	// requestSuccessTotal is prometheus counter for total number of successful requests
 	requestSuccessTotal *prometheus.CounterVec
-	// channel for requeasts to be passed to workers
-	reqChan chan *openaiserverapi.CompletionReqCtx
 }
 
 // LoRAs usage info for requests execution


pkg/llm-d-inference-sim/lora.go
```diff
@@ -70,3 +70,70 @@ func (s *VllmSimulator) unloadLora(ctx *fasthttp.RequestCtx) {
...
+// Checks if the LoRA adaptor is loaded
+func (s *VllmSimulator) loraIsLoaded(model string) bool {
+	if !s.isLora(model) {
+		return true
+	}
+
+	s.loras.mux.RLock()
+	defer s.loras.mux.RUnlock()
+
+	_, ok := s.loras.usedLoras[model]
```
mayabar: consider change name from usedLoras to loadedLoras

pkg/llm-d-inference-sim/lora.go
```diff
@@ -70,3 +70,70 @@ func (s *VllmSimulator) unloadLora(ctx *fasthttp.RequestCtx) {
...
+			}
+		}
+	}
+	if ok {
+		s.loras.usedLoras[model]++
+	}
+	return ok
+}
+
+func (s *VllmSimulator) incrementLora(model string) {
```
mayabar, line 119: please add a comment that this function can be invoked only if the lora is loaded

pkg/llm-d-inference-sim/lora.go
```diff
@@ -70,3 +70,70 @@ func (s *VllmSimulator) unloadLora(ctx *fasthttp.RequestCtx) {
...
+
+func (s *VllmSimulator) decrementLora(model string) {
+	if model == "" || !s.isLora(model) {
+		return
+	}
+
+	s.loras.mux.Lock()
+	defer s.loras.mux.Unlock()
+
+	s.loras.usedLoras[model] -= 1
```
mayabar: s.loras.usedLoras[model]-- ?

pkg/llm-d-inference-sim/worker.go
```diff
@@ -0,0 +1,154 @@
...
+	finishedChan chan *requestCompleted
+	// the request processor
+	processor requestProcessor
+}
+
+func (w *worker) waitForRequests() {
+	for {
+		select {
+		case <-w.ctx.Done():
+			w.logger.V(4).Info("worker done", "id", w.id)
```
mayabar: maybe define some constant for 4 , like in kv-manager DEBUG=4

pkg/llm-d-inference-sim/worker.go
```diff
@@ -0,0 +1,154 @@
...
+	processRequest(reqCtx *openaiserverapi.CompletionReqCtx)
+}
+
+func (s *VllmSimulator) processRequest(reqCtx *openaiserverapi.CompletionReqCtx) {
+	req := reqCtx.CompletionReq
+	model := req.GetModel()
+	displayModel := s.getDisplayedModelName(model)
+
+	// decrement waiting and increment running requests count
+	s.metrics.waitingReqChan <- -1
```
mayabar: consider take care of waiting requests count like loras

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -273,6 +339,101 @@ func (s *VllmSimulator) Printf(format string, args ...interface{}) {
...
+					// send this request for processing in this worker
+					s.logger.V(4).Info("LoRA can be removed: a free worker and request were found - sending to processing",
+						"model", nextReq.CompletionReq.GetModel(),
+						"req", nextReq.CompletionReq.GetRequestID(), "worker", worker.id)
+					worker.reqChan <- nextReq
+				} else {
+					// there are no requests to run (either the queue is empty or maxLoras was reached)
+					s.freeWorkers <- worker
+					break
+				}
```
mayabar: consider to create a function for this

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -273,6 +339,101 @@ func (s *VllmSimulator) Printf(format string, args ...interface{}) {
...
+				s.metrics.lorasChan <- loraUsage{reqCtx.CompletionReq.GetModel(), waitingUsageState}
+				break
+			}
+
+			if s.isLora(model) {
+				// check if lora usage allows the request to run
+				if !s.loadLora(model) {
+					// free the worker, but not directly add it to freeWorkers,
+					// we want to make sure that there is no new request in the queue that can run
+					s.workerFinished <- &requestCompleted{worker: worker, model: ""}
```
mayabar: **Requires additional discussion:**
can we free the worker instead of sending "workerFinished" 
since there is only one select statement

pkg/llm-d-inference-sim/lora.go
irar2: renamed

pkg/llm-d-inference-sim/lora.go
irar2: added

pkg/llm-d-inference-sim/lora.go
irar2: ok

pkg/llm-d-inference-sim/worker.go
irar2: done

pkg/llm-d-inference-sim/simulator.go
irar2: done

pkg/llm-d-inference-sim/simulator.go
irar2: Yes, the previous case was that the request can arrive in parallel, but now they go through this select, so it is safe to just free the worker here.

pkg/llm-d-inference-sim/worker_test.go
```diff
@@ -0,0 +1,613 @@
...
+// after a delay one more request to lora1 (req number 3).
+// All the requests to lora1 should be handled before the requests to lora2.
+// The exact order of the first three requests to lora1 and the requests to lora 2
+// is not important.
+func checkOrderMaxLora1Workers5(orderOfRequests []int) {
+	Expect(orderOfRequests).To(HaveLen(8))
+	for i, reqNum := range orderOfRequests {
+		switch {
+		case i < 3:
+			Expect(reqNum).To(BeNumerically("<", 4))
```
mayabar: <3
same as checkOrderMaxLora5Workers1

pkg/llm-d-inference-sim/worker_test.go
```diff
@@ -0,0 +1,613 @@
...
+// The requests are sent with delays to make sure they enter the queue
+// in the order they are sent.
+// The order of the requests is:
+// 0-lora1 1-lora1 2-lora2 3-lora3 4-lora4 5-lora1 6-lora2 7-lora3 8-lora4
+// The expected order of processing:
+// 015263748
+func checkOrderMaxLora1Workers1(orderOfRequests []int) {
+	expected := []int{0, 1, 5, 2, 6, 3, 7, 4, 8}
+	Expect(orderOfRequests).To(HaveLen(9))
+	for i, reqNum := range orderOfRequests {
```
mayabar: without loop?

pkg/llm-d-inference-sim/worker_test.go
```diff
@@ -0,0 +1,613 @@
...
+// Check the order of requests sent in specific order with max loras = 3
+// The requests are sent with delays to make sure they enter the queue
+// in the order they are sent.
+// The order of the requests is:
+// 0-lora1 1-lora1 2-lora2 3-lora3 4-lora4 5-lora5
+// 6-lora1 7-lora2 8-lora3 9-lora4 10-lora5
+// The expected order of processing:
+// 0, 1, 2, 3, 6, 7, 8, {4, 9}, {5, 10} - the order inside the brackets doesn't matter
+func checkOrderMaxLora3(orderOfRequests []int) {
+	expected1 := []int{0, 1, 2, 3, 6, 7, 8, 4, 9, 5, 10}
```
mayabar: in case of 4 workers - 4 and 9 have specific order

pkg/llm-d-inference-sim/worker_test.go
```diff
@@ -0,0 +1,613 @@
...
+			}
+
+			client, err := startServerWithArgs(ctx, common.ModeRandom, args, nil)
+			Expect(err).NotTo(HaveOccurred())
+
+			openaiclient := openai.NewClient(
+				option.WithBaseURL(baseURL),
+				option.WithHTTPClient(client))
+
+			// Run 2000 requests for 5 loras simultaneously
```
mayabar: 2 loras

pkg/llm-d-inference-sim/worker_test.go
irar2: done

pkg/llm-d-inference-sim/worker_test.go
irar2: fixed

pkg/llm-d-inference-sim/worker_test.go
irar2: done

mayabar (approved):

Issue #213, opened 2025-10-16
feat: Log probabilities support


ruivieira:
**What would you like to be added**:

Log probabilities support for llm-d-inference-sim.

llm-d-inference-sim currently lacks support for log probabilities (logprobs) in completion responses, which limits its usability in integration testing scenarios with evaluation frameworks and tools that require this API feature to be present.

**Why is this needed**:

1. Model Evaluations 

As an example, The lm-evaluation-harness framework uses logprobs extensively for benchmark tasks:

* Multiple-choice tasks (MMLU, HellaSwag, ARC, etc.): These tasks require logprobs to calculate which answer choice has the highest likelihood. Without logprobs, these evaluations cannot run.
* Perplexity measurements: Many benchmarks compute perplexity by summing log probabilities across tokens.
* Loglikelihood scoring: Tasks need to compare the probability of different continuations to select the most likely one.

2. API Coverage
vLLM's OpenAI-compatible API includes logprobs support as specified in the OpenAI API specification:

Text completions: `/v1/completions` endpoint supports logprobs parameter (integer 0-5)
Chat completions: `/v1/chat/completions` endpoint supports `logprobs` boolean and `top_logprobs` integer parameters

API Format:

Text Completion Request (OpenAI format)
```json
{
  "model": "Qwen/Qwen2-0.5B",
  "prompt": "The capital of France is",
  "max_tokens": 5,
  "logprobs": 2
}
```

Text Completion Response (with logprobs)

```json
{
  "id": "cmpl-abc123",
  "object": "text_completion",
  "created": 1234567890,
  "model": "Qwen/Qwen2-0.5B",
  "choices": [
    {
      "text": " Paris",
      "index": 0,
      "logprobs": {
        "tokens": [" Paris"],
        "token_logprobs": [-0.15],
        "top_logprobs": [
          {
            " Paris": -0.15,
            " paris": -2.8
          }
        ],
        "text_offset": [0]
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 5,
    "completion_tokens": 1,
    "total_tokens": 6
  }
}
```

Chat Completion Request (OpenAI format)

```json
{
  "model": "Qwen/Qwen2-0.5B",
  "messages": [
    {"role": "user", "content": "What is 2+2?"}
  ],
  "logprobs": true,
  "top_logprobs": 3
}
```

Chat Completion Response (with logprobs)

```json
{
[...]

PR #215, opened 2025-10-16
feat: Log probabilities support (with cache)

ruivieira:
This PR is a proposal to add log probabilities (logprobs) support to llm-d-inference-sim following vLLM's architecture.

*vLLM-Style LogprobsProcessor*

Added a new `LogprobsProcessor` that following vLLM's implementation (same tokens always produce same logprobs) 

*API coverage*

- `/v1/completions`: `logprobs` parameter (0-5)
- `/v1/chat/completions`: `logprobs` boolean + `top_logprobs` integer
- Per-token logprobs in SSE chunks
- Response format matching vLLM/OpenAI

Refer to https://github.com/llm-d/llm-d-inference-sim/issues/213


pkg/llm-d-inference-sim/logprobs_processor_test.go
```diff
@@ -0,0 +1,216 @@
...
+		t.Errorf("Expected same topLogprobs length, got %d and %d", len(topLogprobs1), len(topLogprobs2))
+	}
+	if len(topLogprobs1) != topK {
+		t.Errorf("Expected topLogprobs length %d, got %d", topK, len(topLogprobs1))
+	}
+
+	// Check cache stats
+	hits, misses, hitRate := processor.GetCacheStats()
+	if hits != 1 {
+		t.Errorf("Expected 1 cache hit, got %d", hits)
```
mayabar: In all tests in this project we use Expect function of onsi/gomega. E.g. Expect(err).NotTo(HaveOccurred()), Expect(totalBlocks).To(Equal(unusedBlocks))...
Can you please change to be consistent with other tests.

mayabar: In all tests in this project we use Expect function of onsi/gomega. E.g. Expect(err).NotTo(HaveOccurred()), Expect(totalBlocks).To(Equal(unusedBlocks))...
Can you please change to be consistent with other tests.

pkg/llm-d-inference-sim/streaming.go
```diff
@@ -128,8 +129,16 @@ func (s *VllmSimulator) sendTokenChunks(context *streamingContext, w *bufio.Writ
 		if i == len(genTokens)-1 && (finishReason == dataset.LengthFinishReason || finishReason == dataset.ToolsFinishReason) {
 			finishReasonToSend = &finishReason
 		}
+
```
mayabar, line 135: this part does not change during the loop, could be moved to be calculated before the loop starts

pkg/llm-d-inference-sim/streaming.go
```diff
@@ -201,7 +210,7 @@ func (s *VllmSimulator) createTextCompletionChunk(context *streamingContext, tok
 // createChatCompletionChunk creates and returns a CompletionRespChunk, a single chunk of streamed completion
 // API response, for chat completion. It sets either role, or token, or tool call info in the message.
 func (s *VllmSimulator) createChatCompletionChunk(context *streamingContext, token string, tool *openaiserverapi.ToolCall,
-	role string, finishReason *string) openaiserverapi.CompletionRespChunk {
+	role string, finishReason *string, includeLogprobs bool, topK int) openaiserverapi.CompletionRespChunk {
```
mayabar, line 216: The boolean value which defines if logprob should be returned and the topk value are constant during a request processing, such kind of values we store in streamingContext.
If token is empty no need to create all logprob related fields, otherwise relevant response part will be created.

mayabar:
@ruivieira Thank you for your contribution!
I have added some general comments regarding the project's conventions for now. 
I will perform a more detailed review of the PR soon.

pkg/llm-d-inference-sim/streaming.go
ruivieira: Good point! Moved to outside.

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -393,6 +396,8 @@ func (s *VllmSimulator) reqProcessingWorker(ctx context.Context, id int) {
 							doRemotePrefill:     req.IsDoRemotePrefill(),
 							nPromptTokens:       usageData.PromptTokens,
 							nCachedPromptTokens: reqCtx.CompletionReq.GetNumberOfCachedPromptTokens(),
+							requestID:           req.GetRequestID(),
+							request:             req,
```
irar2, line 400: Please add shouldIncludeLogprobs and topK (maybe change the name to explain the context) to streamingContext instead of request.

(And thanks for adding the missing request id)

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -490,7 +529,7 @@ func (s *VllmSimulator) createCompletionResponse(isChatCompletion bool, respToke
 // usageData - usage (tokens statistics) for this response
 func (s *VllmSimulator) sendResponse(reqCtx *openaiserverapi.CompletionReqCtx, respTokens []string, toolCalls []openaiserverapi.ToolCall,
 	modelName string, finishReason string, usageData *openaiserverapi.Usage) {
-	resp := s.createCompletionResponse(reqCtx.IsChatCompletion, respTokens, toolCalls, &finishReason, usageData, modelName,
+	resp := s.createCompletionResponse(reqCtx.CompletionReq, reqCtx.IsChatCompletion, respTokens, toolCalls, &finishReason, usageData, modelName,
```
irar2, line 532: Please pass shouldIncludeLogprobs and topK instead of request

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -427,15 +432,35 @@ func (s *VllmSimulator) responseSentCallback(model string, isChatCompletion bool
...
 // as defined by isChatCompletion
+// req - the original completion request
 // respTokens - tokenized content to be sent in the response
 // toolCalls - tool calls to be sent in the response
 // finishReason - a pointer to string that represents finish reason, can be nil or stop or length, ...
 // usageData - usage (tokens statistics) for this response
 // modelName - display name returned to the client and used in metrics. It is either the first alias
 // from --served-model-name (for a base-model request) or the LoRA adapter name (for a LoRA request).
-func (s *VllmSimulator) createCompletionResponse(isChatCompletion bool, respTokens []string, toolCalls []openaiserverapi.ToolCall,
+func (s *VllmSimulator) createCompletionResponse(req openaiserverapi.CompletionRequest, isChatCompletion bool, respTokens []string, toolCalls []openaiserverapi.ToolCall,
```
irar2, line 463: Please pass shouldIncludeLogprobs and topK instead of request

pkg/llm-d-inference-sim/logprobs_processor.go
```diff
@@ -0,0 +1,233 @@
...
+	"sync"
+
+	openaiserverapi "github.com/llm-d/llm-d-inference-sim/pkg/openai-server-api"
+)
+
+// LogprobData represents cached logprob information for a token
+type LogprobData struct {
+	MainLogprob float64                                 `json:"main_logprob"`
+	TopLogprobs []openaiserverapi.ChatCompletionLogProb `json:"top_logprobs"`
+}
```
irar2, line 31: Why do the struct and the fields have to be public? And why are the json tags needed?

pkg/llm-d-inference-sim/logprobs_processor_test.go
```diff
@@ -0,0 +1,162 @@
...
+	Expect(misses).To(Equal(int64(3)))
+
+	// Access one of the earlier tokens - may or may not be in cache due to eviction
+	processor.GetLogprobs("token1", 1)
+
+	// Cache should be working (some entries may have been evicted)
+	hits2, misses2, _ := processor.GetCacheStats()
+	Expect(hits2).To(BeNumerically(">=", 0))
+	Expect(misses2).To(BeNumerically(">=", 3))
+}
```
irar2, line 162: Could you please add tests for the simulator with logprobs:
completions and chat completions requests with logprobs with and without streaming, and check the response?

pkg/llm-d-inference-sim/logprobs_processor.go
```diff
@@ -0,0 +1,233 @@
...
+
+// LogprobsProcessor handles logprobs generation and caching following vLLM architecture
+type LogprobsProcessor struct {
+	// tokenCache caches logprobs by token content and topK to avoid recomputation
+	tokenCache map[string]*LogprobData
+	cacheMutex sync.RWMutex
+
+	// cacheHits and cacheMisses for metrics
+	cacheHits   int64
```
mayabar, line 40: cacheHits and cacheMisses should be changed under protection of the mutex or be an atomic counters (from sync/atomic package)

pkg/llm-d-inference-sim/logprobs_processor.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+}
+
+// generateCacheKey creates a deterministic key for caching based on token and topK
+func (lp *LogprobsProcessor) generateCacheKey(token string, topK int) string {
+	return fmt.Sprintf("%s:%d", token, topK)
+}
+
+// generateDeterministicLogprobs creates logprobs with deterministic values based on token content
+// This follows vLLM's approach of consistent logprobs for the same token in similar contexts
```
mayabar, line 65: @ruivieira can you please explain how context is taken into consideration here. As I understand a token should give same logprob only in the same/similar context. What does context means here? Isn't it the previous tokens? But they are not passed to this function.

pkg/llm-d-inference-sim/logprobs_processor.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// Store in cache (with size limit)
+	lp.cacheMutex.Lock()
+	if len(lp.tokenCache) >= lp.maxCacheSize {
+		// Simple eviction: remove oldest entry
+		// In production, this could use LRU or other strategies
+		for k := range lp.tokenCache {
+			delete(lp.tokenCache, k)
+			break
+		}
```
mayabar, line 135: Iteration on a map will not be in order of adding objects to the map, is we want to remove the oldest item - need to store the insertion date/time or latest usage date/time

pkg/llm-d-inference-sim/logprobs_processor.go
```diff
@@ -0,0 +1,233 @@
...
+	hitRate = 0.0
+	if total > 0 {
+		hitRate = float64(lp.cacheHits) / float64(total)
+	}
+
+	return lp.cacheHits, lp.cacheMisses, hitRate
+}
+
+// ClearCache clears the logprobs cache
+func (lp *LogprobsProcessor) ClearCache() {
```
mayabar, line 226: Not used

pkg/llm-d-inference-sim/logprobs_processor.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+
+		textOffset += len(token)
+	}
+
+	return logprobs
+}
+
+// GetCacheStats returns cache performance statistics
+func (lp *LogprobsProcessor) GetCacheStats() (hits, misses int64, hitRate float64) {
```
mayabar, line 212: consider to not add function used in test only to the class, we may add an additional file or maybe choose not to use it in test

ruivieira:
@mayabar That's a good point. My original implementation did take that approach, but ultimately I've pushed this one mainly for reproducibility and consistent (even if simulated) results.

I agree that purely for API coverage generating random responses is enough and avoids the overhead. I'll open a new PR with a new implementation and, if preferred, we close this one?

ruivieira:
@mayabar did not see your comment before opening https://github.com/llm-d/llm-d-inference-sim/pull/221 :)
Let me know if you still prefer to change this one, fine for me either way. Thanks!

PR #221, opened 2025-10-20
feat: Log probabilities support

ruivieira:
This PR implements log probabilities support in llm-d-inference-sim enabling compatibility with evaluation frameworks and other API clients that require token-level probability.

_API Coverage_

- `/v1/completions`: logprobs parameter with `TextLogprobs` response structure
- `/v1/chat/completions`: logprobs boolean + top_logprobs integer with `ChatLogprobs` response structure
- Response format matching vLLM/OpenAI

Refer to https://github.com/llm-d/llm-d-inference-sim/issues/213
Replaces https://github.com/llm-d/llm-d-inference-sim/pull/215

### DIFF
--- a/pkg/common/logprobs_test.go
+++ b/pkg/common/logprobs_test.go
@@ -1,0 +1,253 @@
+/*
+Copyright 2025 The llm-d-inference-sim Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Logprobs", func() {
+
+	Context("GenerateTextLogprobs", func() {
+		It("should generate correct text logprobs structure", func() {
+			tokens := []string{" Paris", ",", " the", " capital"}
+			logprobsCount := 2
+
+			logprobs := GenerateTextLogprobs(tokens, logprobsCount)
+
+			Expect(logprobs).NotTo(BeNil())
+			Expect(logprobs.Tokens).To(HaveLen(len(tokens)))
+			Expect(logprobs.TokenLogprobs).To(HaveLen(len(tokens)))
+			Expect(logprobs.TopLogprobs).To(HaveLen(len(tokens)))
+			Expect(logprobs.TextOffset).To(HaveLen(len(tokens)))
+
+			// Check that each top logprobs entry has the expected number of alternatives
+			for i, topLogprob := range logprobs.TopLogprobs {
+				Expect(topLogprob).To(HaveLen(logprobsCount))
+				// Check that the main token is included in the alternatives
+				Expect(topLogprob).To(HaveKey(tokens[i]))
+			}
+
+			// Check text offsets are calculated correctly (byte-based)
+			expectedOffsets := []int{0, 6, 7, 11} // " Paris" - 6, "," - 1, " the" -4, " capital" - 11
+			for i, expected := range expectedOffsets {
+				Expect(logprobs.TextOffset[i]).To(Equal(expected))
+			}
+
+			// Check deterministic logprobs
+			expectedLogprob0 := -1.0 // defaultLogprob - float64(0%3)*0.1
+			Expect(logprobs.TokenLogprobs[0]).To(Equal(expectedLogprob0))
+		})
+	})
+
+	Context("GenerateChatLogprobs", func() {
+		It("should generate correct chat logprobs structure", func() {
+			tokens := []string{"4"}
+			topLogprobsCount := 3
+
+			logprobs := GenerateChatLogprobs(tokens, topLogprobsCount)
+
+			Expect(logprobs).NotTo(BeNil())
+			Expect(logprobs.Content).To(HaveLen(len(tokens)))
+
+			content := logprobs.Content[0]
+			Expect(content.Token).To(Equal(tokens[0]))
+			Expect(content.Bytes).To(HaveLen(len(tokens[0])))
+			Expect(content.TopLogprobs).To(HaveLen(topLogprobsCount))
+
+			// Check that the main token is the first in top logprobs
+			Expect(content.TopLogprobs[0].Token).To(Equal(tokens[0]))
+
+			// Check alternative tokens follow the pattern
+			expectedAlt1 := "4_1"
+			Expect(content.TopLogprobs[1].Token).To(Equal(expectedAlt1))
+
+			// Check byte conversion
+			expectedBytes := []int{52} // byte value of '4'
+			for i, expected := range expectedBytes {
+				Expect(content.Bytes[i]).To(Equal(expected))
+			}
+
+			// Check deterministic logprobs
+			expectedLogprob := -1.0 // defaultLogprob - float64(0%3)*0.1
+			Expect(content.Logprob).To(Equal(expectedLogprob))
+		})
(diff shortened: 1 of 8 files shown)


pkg/llm-d-inference-sim/simulator.go
```diff
@@ -439,13 +439,14 @@ func (s *VllmSimulator) responseSentCallback(model string, isChatCompletion bool
...
 // as defined by isChatCompletion
+// req - the completion request context containing logprobs settings
 // respTokens - tokenized content to be sent in the response
 // toolCalls - tool calls to be sent in the response
 // finishReason - a pointer to string that represents finish reason, can be nil or stop or length, ...
 // usageData - usage (tokens statistics) for this response
 // modelName - display name returned to the client and used in metrics. It is either the first alias
 // from --served-model-name (for a base-model request) or the LoRA adapter name (for a LoRA request).
-func (s *VllmSimulator) createCompletionResponse(isChatCompletion bool, respTokens []string, toolCalls []openaiserverapi.ToolCall,
+func (s *VllmSimulator) createCompletionResponse(req openaiserverapi.CompletionRequest, isChatCompletion bool, respTokens []string, toolCalls []openaiserverapi.ToolCall,
```
mayabar: Instead of passing request, please pass a pointer to the logprob data, in case there is no need to add logprobs to the response (req.IncludeLogprobs() && toolCalls == nil ) you can pass nil

pkg/openai-server-api/logprobs_utils.go
```diff
@@ -0,0 +1,150 @@
...
+const (
+	// Default logprob value
+	defaultLogprob = -1.0
+)
+
+// NOTE: These functions produce synthetic data for API shape compatibility.
+// The logprobs are deterministic placeholders and have no semantic meaning.
+
+// GenerateTextLogprobs generates synthetic log probabilities for text completion responses
+func GenerateTextLogprobs(tokens []string, logprobsCount int) *TextLogprobs {
```
mayabar: Maybe it's possible to create some helper function which will contains the "algorithm" of probability calculation (main token probability with cycle of 3 and additional tokens probabilities going down by 0.5)

pkg/openai-server-api/logprobs_utils.go
```diff
@@ -0,0 +1,150 @@
+/*
```
mayabar: please move this file to common

pkg/openai-server-api/request.go
```diff
@@ -253,6 +265,19 @@ func (req *ChatCompletionRequest) GetFullPrompt() string {
 	return prompt
 }
 
+func (c *ChatCompletionRequest) GetLogprobs() *int {
+	// For chat completions, this method returns nil
+	return nil
```
mayabar: Why getter for logprobs value is handled differently in text completions and chat completions?
ChatCompletionRequest returns nil in GetLogprobs, on the other hand, TextCompletionRequest returns value in GetLogprobs GetTopLogprobs.
**Additional suggestion**, we may have only one "GetLogprobs() *int" function in the interface, it will check if logprobs should be returned and return the value if needed, returning nil will mean that logprobs should not be included in the response. Each kind of completion will do it specific logic.

pkg/common/logprobs_test.go
```diff
@@ -0,0 +1,216 @@
...
+			Expect(logprobs.Tokens).To(HaveLen(len(tokens)))
+			Expect(logprobs.TokenLogprobs).To(HaveLen(len(tokens)))
+			Expect(logprobs.TopLogprobs).To(HaveLen(len(tokens)))
+			Expect(logprobs.TextOffset).To(HaveLen(len(tokens)))
+
+			// Check that each top logprobs entry has the expected number of alternatives
+			for i, topLogprob := range logprobs.TopLogprobs {
+				Expect(topLogprob).To(HaveLen(logprobsCount))
+				// Check that the main token is included in the alternatives
+				Expect(topLogprob).To(HaveKey(tokens[i]))
```
mayabar, line 43: Consider to add a check that probabilities going lower as i is growing

mayabar (changes requested):
@ruivieira I added some in-place comments 
In addition, please add logprobs support for requests with streaming=true (sendStreamingResponse), please take into consideration comments in PR #215 relevant to streaming. Please add test for streaming scenarios.

pkg/llm-d-inference-sim/simulator.go
ruivieira: Implemented in https://github.com/llm-d/llm-d-inference-sim/pull/221/commits/e98ded1c0c4f3ed6f0c523255fe904285443b928

pkg/openai-server-api/logprobs_utils.go
ruivieira: Implemented in https://github.com/llm-d/llm-d-inference-sim/pull/221/commits/e98ded1c0c4f3ed6f0c523255fe904285443b928

pkg/openai-server-api/logprobs_utils.go
ruivieira: Implemented in https://github.com/llm-d/llm-d-inference-sim/pull/221/commits/e98ded1c0c4f3ed6f0c523255fe904285443b928

Since there are some cross-package imports, I've opted to call the types from `common` (`common.ChatLogprobs`) into `openaiserverapi` rather than the other way around. Please let me know if you prefer the other way around, or another refactor.

pkg/openai-server-api/request.go
ruivieira: Unified interface in https://github.com/llm-d/llm-d-inference-sim/pull/221/commits/e98ded1c0c4f3ed6f0c523255fe904285443b928

pkg/common/logprobs_test.go
ruivieira: Implemented in https://github.com/llm-d/llm-d-inference-sim/pull/221/commits/e98ded1c0c4f3ed6f0c523255fe904285443b928

pkg/openai-server-api/logprobs_utils.go
mayabar: The dependency between openaiserverapi and common was in the wrong direction. Ira fixed this in PR #224 
Now it will be possible to move the logprobs_utils to common.

pkg/llm-d-inference-sim/streaming.go
```diff
@@ -38,6 +38,10 @@ type streamingContext struct {
 	nPromptTokens       int
 	nCachedPromptTokens int
 	requestID           string
+	// Logprobs configuration
+	includeLogprobs bool
+	topLogprobs     int
+	logprobsCount   *int // Text completions
```
mayabar: there is a LogprobData type which contains the same fields, please consider to use it

pkg/openai-server-api/request.go
```diff
@@ -72,6 +72,14 @@ type CompletionRequest interface {
...
 	// for text completion - max_tokens field is used
 	ExtractMaxTokens() *int64
+	// GetLogprobs returns the logprobs parameter (for text completions)
+	GetLogprobs() *int
+	// IncludeLogprobs returns true if logprobs should be included (for chat completions)
+	IncludeLogprobs() bool
+	// GetTopLogprobs returns the number of top logprobs to include (for chat completions)
+	GetTopLogprobs() *int
+	// GetTopLogprobsCount returns the computed count of top logprobs to include
+	GetTopLogprobsCount() int
```
mayabar: @ruivieira could you please explain why we need all 4 functions?
What about having just `GetLogprobs() *int`, which will be implemented differently for completions and chat completions, where each version uses the relevant fields of the request.
In all cases, we only need to know if logprobs should be included in the response and how many options to add. So a single function which returns a pointer to int isshoul be enough - in case there is no need in logprobls - nil , in case logprobs are required - we will have the number of options to include.
What do you think?

mayabar (commented):
Hi @ruivieira , I added some comments. 
Please pay attention that there are changes in tests utility functions.

pkg/openai-server-api/request.go
ruivieira: @mayabar I agree, this is a much better design. Implemented it and that also closes https://github.com/llm-d/llm-d-inference-sim/pull/221#discussion_r2469243619.

mayabar (approved):

PR #250, opened 2025-11-03
Support vllm:max_num_generation_tokens metrics

### LINKED ISSUE #243: Add vllm:request_max_num_generation_tokens metric
vllm:request_max_num_generation_tokens - This is the minimum of max-model-len - prompt length and max_tokens if defined.

mayabar:
Currently `vllm:max_num_generation_tokens` report same values as `vllm:request_generation_tokens` since response always contains only one choice.

Fixes #243

### DIFF
--- a/pkg/common/config_test.go
+++ b/pkg/common/config_test.go
@@ -435,12 +435,12 @@ var _ = Describe("Simulator configuration", func() {
 		{
 			name:          "invalid (negative) zmq-max-connect-attempts for argument",
 			args:          []string{"cmd", "--zmq-max-connect-attempts", "-1", "--config", "../../manifests/config.yaml"},
-			expectedError: "zmq retries times cannot be more negative",
+			expectedError: "zmq retries times cannot be negative",
 		},
 		{
 			name:          "invalid (negative) zmq-max-connect-attempts for config file",
 			args:          []string{"cmd", "--config", "../../manifests/invalid-config.yaml"},
-			expectedError: "zmq retries times cannot be more negative",
+			expectedError: "zmq retries times cannot be negative",
 		},
 		{
 			name: "invalid (negative) prefill-overhead",
--- a/pkg/common/config_test.go
+++ b/pkg/common/config_test.go
@@ -307,8 +307,8 @@ var _ = Describe("Simulator configuration", func() {
 		{
 			name: "invalid lora",
 			args: []string{"cmd", "--config", "../../manifests/config.yaml",
-				"--lora-modules", "[{\"path\":\"/path/to/lora15\"}]"},
-			expectedError: "json: cannot unmarshal array into Go value of type common.LoraModule",
+				"--lora-modules", "{\"path\":\"/path/to/lora15\"}"},
+			expectedError: "empty LoRA name",
 		},
 		{
 			name:          "invalid max-model-len",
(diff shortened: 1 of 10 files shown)


pkg/common/config_test.go
```diff
@@ -283,212 +285,263 @@ var _ = Describe("Simulator configuration", func() {
...
-			args: []string{"cmd", "--mode", "hello", "--config", "../../manifests/config.yaml"},
+			name:          "invalid mode",
+			args:          []string{"cmd", "--mode", "hello", "--config", "../../manifests/config.yaml"},
+			expectedError: "invalid mode ",
 		},
 		{
 			name: "invalid lora",
 			args: []string{"cmd", "--config", "../../manifests/config.yaml",
 				"--lora-modules", "[{\"path\":\"/path/to/lora15\"}]"},
+			expectedError: "json: cannot unmarshal array into Go value of type common.LoraModule",
```
irar2: This is an error in the test, we want to check here that the name is not empty, so please remove the brackets: `"--lora-modules", "{\"path\":\"/path/to/lora15\"}"}`, and expect `empty LoRA name` error

pkg/common/config_test.go
```diff
@@ -283,212 +285,263 @@ var _ = Describe("Simulator configuration", func() {
...
+			name:          "invalid (negative) zmq-max-connect-attempts for argument",
+			args:          []string{"cmd", "--zmq-max-connect-attempts", "-1", "--config", "../../manifests/config.yaml"},
+			expectedError: "for \"--zmq-max-connect-attempts\" flag: strconv.ParseUint: parsing",
 		},
 		{
-			name: "invalid (negative) zmq-max-connect-attempts for config file",
-			args: []string{"cmd", "--config", "../../manifests/invalid-config.yaml"},
+			name:          "invalid (negative) zmq-max-connect-attempts for config file",
+			args:          []string{"cmd", "--config", "../../manifests/invalid-config.yaml"},
+			expectedError: "failed to unmarshal configuration: yaml: unmarshal errors:",
```
irar2: Maybe we should change the type to int to get better errors? What do you think?

pkg/common/utils.go
```diff
@@ -161,3 +162,16 @@ func WriteToChannel[T any](channel chan T, object T, logger logr.Logger, channel
 		logger.V(logging.WARN).Info("failed to write to", "channel", channelName)
 	}
 }
+
+func MaxIntSlice(numbers []int) (int, error) {
```
irar2, line 168: Please add a comment describing what this function does

pkg/llm-d-inference-sim/metrics.go
```diff
@@ -656,8 +675,12 @@ func (s *VllmSimulator) recordRequestUpdater(ctx context.Context) {
 type requestSuccessEvent struct {
 	// promptTokens is the number of input (prompt) tokens in the request
 	promptTokens int
-	// generationTokens is the number of generated (output) tokens in the response
+	// generationTokens is the number of generated (output) tokens in the response,
+	// in case of response with multiple choices contains sum of lengths of all choises
```
irar2: typo in choises

irar2 (changes requested):

pkg/common/config_test.go
mayabar: done

pkg/common/config_test.go
mayabar: Changed to int

irar2 (approved):

Issue #251, opened 2025-11-05
Does vLLM-Simulator support Prefill/Decode separation simulation?


googs1025:
Does the current **vLLM-Simulator** support **Prefill/Decode separation** simulation?  
If yes, could we provide an **official example** (YAML ) so that we can quickly verify the P/D path in CI/local development.

googs1025:
some info: https://github.com/llm-d/llm-d-inference-scheduler/blob/main/deploy/components/vllm-sim-pd/deployments.yaml

But I'm not sure if this example already use PD separation.

googs1025:
@shmuelk Thank you for your reply. 😄  I tested it locally and it works. In addition, I would like to request that a similar example YAML be provided in [manifests](https://github.com/llm-d/llm-d-inference-sim/tree/main/manifests) section of the llm-d-inference-sim project. I plan to add a YAML file and a similar README guide.

PR #253, opened 2025-11-09
docs: add P/D disaggregation example in manifests/disaggregation

googs1025:
- docs: add P/D disaggregation example in manifests/disaggregation

fix: https://github.com/llm-d/llm-d-inference-sim/issues/251#issuecomment-3507535574

### DIFF
--- a/manifests/disaggregation/README.md
+++ b/manifests/disaggregation/README.md
@@ -1,0 +1,83 @@
+## Prefill/Decode Disaggregation Deployment Guide
+
+This guide demonstrates how to deploy the LLM Disaggregation Simulator (llm-d-sim) in a Kubernetes cluster using a separated Prefill and Decode (P/D) architecture. 
+
+The [`routing-sidecar`](https://github.com/llm-d/llm-d-routing-sidecar) runs alongside the Decode service and acts as a reverse proxy: it receives client requests, forwards the prefill phase to a dedicated Prefill service (based on the x-prefiller-host-port header), and then handles the decode phase locally.
+
+> This is a standalone simulation setup, intended for testing and validating P/D workflows without requiring the [llm-d-inference-scheduler](https://github.com/llm-d/llm-d-inference-scheduler). 
+> It uses standard Kubernetes Services for internal communication between components.
+
+### Quick Start
+
+1. Deploy the Application
+   Apply the provided manifest (e.g., vllm-sim-pd.yaml) to your Kubernetes cluster:
+
+```bash
+kubectl apply -f vllm-sim-pd.yaml
+```
+
+> This manifest defines two Deployments (vllm-sim-p for Prefill, vllm-sim-d for Decode) and two Services for internal and external communication.
+
+2. Verify Pods Are Ready
+   Check that all pods are running:
+
+```bash
+kubectl get pods -l 'llm-d.ai/role in (prefill,decode)'
+```
+
+Expected output:
+
+```bash
+NAME                          READY   STATUS    RESTARTS   AGE
+vllm-sim-d-685b57d694-d6qxg   2/2     Running   0          12m
+vllm-sim-p-7b768565d9-79j97   1/1     Running   0          12m
+```
+
+### Send a Disaggregated Request Using kubectl port-forward
+To access both the Decode services from your local machine, use `kubectl port-forward` to forward their ports to your localhost.
+
+### Forward the Decode Service Port
+Open a terminal and run:
+
+```bash
+kubectl port-forward svc/vllm-sim-d-service 8000:8000
+```
+
+This command forwards port 8000 from the `vllm-sim-d-service` to your local machine's port 8000.
+
+#### Test the Disaggregated Flow
+
+Now, send a request to the forwarded Decode service port with the necessary headers:
+
+```bash
+curl -v http://localhost:8000/v1/chat/completions \
+  -H "Content-Type: application/json" \
+  -H "x-prefiller-host-port: vllm-sim-p-service:8000" \
+  -d '{
+    "model": "meta-llama/Llama-3.1-8B-Instruct",
+    "messages": [{"role": "user", "content": "Hello from P/D architecture!"}],
+    "max_tokens": 32
+  }'
+```
+
+>  Critical Header:
+>```
+>x-prefiller-host-port: vllm-sim-p-service:8000
+>```
+>This header must be provided by the client in standalone mode. It tells the `routing-sidecar` where to send the prefill request. The value should be a Kubernetes Service name + port (or any resolvable `host:port` reachable from the sidecar pod).
+>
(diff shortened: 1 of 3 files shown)


manifests/disaggregation/vllm-sim-pd.yaml
```diff
@@ -0,0 +1,99 @@
+---
+# Prefill Deployment
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: vllm-sim-p
+  labels:
+    app: my-llm-pool
```
mayabar: why the "my-llm-pool" label is required? when the simulator is installed as part of llm-d - it defines the innference-pool, but here this is a stand-alone example

manifests/disaggregation/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+This command forwards port 8000 from the `vllm-sim-d-service` to your local machine's port 8000.
+
+#### Test the Disaggregated Flow
+
+Now, send a request to the forwarded Decode service port with the necessary headers:
+
+```bash
+curl -v http://localhost:8001/v1/chat/completions \
```
mayabar: port should be 8000

manifests/disaggregation/vllm-sim-pd.yaml
```diff
@@ -0,0 +1,99 @@
...
+      app: my-llm-pool
+  template:
+    metadata:
+      labels:
+        app: my-llm-pool
+        llm-d.ai/role: decode
+    spec:
+      containers:
+        - name: routing-sidecar
+          image: ghcr.io/llm-d/llm-d-routing-sidecar:latest
```
mayabar: there is no latest tag for side-car

manifests/disaggregation/README.md
googs1025: oh, thanks for reminding

manifests/disaggregation/vllm-sim-pd.yaml
googs1025: removed, only use `llm-d.ai/role` label

manifests/disaggregation/vllm-sim-pd.yaml
googs1025: Yes, that's also where I have some questions. When I was testing, I used `ghcr.io/llm-d/llm-d-routing-sidecar:v0.3.1-rc.1`, which I referenced here: https://github.com/llm-d/llm-d-routing-sidecar/releases.


However, I see "latest" being used here, which I assumed was due to network issues I can't downloads.

https://github.com/llm-d/llm-d-inference-scheduler/blob/091f07e7809160bc062be2199500f2675e9cf734/deploy/components/vllm-sim-pd/deployments.yaml#L74

googs1025:
> @googs1025 I tested the configuration from the scheduler and turned out that there is a new version of nixl connector and simulator should be changed to by in-sync with it. Created an issue for this #255

change to `- "--connector=nixlv2"`

googs1025:
> > can we move on this pr now? 😄 @mayabar
> 
> Yes, sure
> 
> Can you please explain me the purpose of this installation yaml? On the one hand it contains the sidecar, so seems that this is part of simulator usage in context of llm-d-scheduler, on the other hand the sample is in the simulator, so seems that it's a stand alone. If this is stand-alone scenario - please add explanation about the 'x-prefiller-host-port' header, which is required for sidecar.
> 
> General comment, if all latency parameters are 0 - it actually does not present the real scenario. If you define a high value of prefill time per token and low time for prefill transfer per token, this will give more realistic behavior of the decode pod.


thank you for the thoughtful feedback! @mayabar 

### Purpose of This Installation YAML

The purpose of this installation YAML is to provide a **standalone simulation** of the P/D disaggregation, **without requiring the full `llm-d-inference-scheduler`**. We put this in the `simulator/` directory precisely because it’s meant to be self-contained and easy to run for testing.

In this setup:
- The **Prefill service** (`vllm-sim-p`) runs independently.
- The **Decode pod** includes two containers:
  - `vllm-decode`: simulates the decode phase,
  - `routing-sidecar`: acts as a reverse proxy that forwards prefill requests based on the `x-prefiller-host-port` header.

> 🔑 **Important**: Since there’s no scheduler or gateway to inject routing info automatically, **the client must include the `x-prefiller-host-port` header** in every request (e.g., `x-prefiller-host-port: vllm-sim-p-service:8000`). I’ll update the guide to make this requirement explicit.

### Realistic Config

To better simulate actual P/D performance:

- We can set a **`--prefill-time-per-token`** and **`--prefill-transfer-time-per-token`** .

like this:
```bash
--prefill-time-per-token=20 \
--prefill-time-std-dev=3 \
```


### Additional Context

It's worth noting that **this is the first complete example of P/D disaggregation in the `llm-d` project**, filling an important gap in our documentation and examples. 

Thanks again for the great suggestions!

mayabar:
@googs1025 Great explanation! Please go ahead with the changes you mentioned above, thanks a lot!

manifests/disaggregation/README.md
```diff
@@ -0,0 +1,83 @@
...
+>  Critical Header:
+>```
+>x-prefiller-host-port: vllm-sim-p-service:8000
+>```
+>This header must be provided by the client in standalone mode. It tells the `routing-sidecar` where to send the prefill request. The value should be a Kubernetes Service name + port (or any resolvable `host:port` reachable from the sidecar pod).
+>
+> In production deployments using `llm-d-inference-scheduler`, this header is typically injected automatically by the scheduler or gateway—but in this standalone simulator, the client must set it explicitly.
+
+
+#### Realistic Config
```
googs1025, line 72: add this part to tell user how to use config

manifests/disaggregation/vllm-sim-pd.yaml
```diff
@@ -0,0 +1,93 @@
...
+      containers:
+        - name: vllm-prefill
+          image: ghcr.io/llm-d/llm-d-inference-sim:latest
+          imagePullPolicy: IfNotPresent
+          args:
+            - "--v=4"
+            - "--port=8000"
+            - "--model=meta-llama/Llama-3.1-8B-Instruct"
+            - "--data-parallel-size=1"
+            - "--prefill-time-per-token=200"
```
googs1025, line 26: add flag

googs1025:
@mayabar done 😄 
plz check when you have time!

manifests/disaggregation/vllm-sim-pd.yaml
mayabar: The tag used in the installation process can be defined using VLLM_SIMULATOR_TAG - see https://github.com/llm-d/llm-d-inference-scheduler/blob/091f07e7809160bc062be2199500f2675e9cf734/deploy/components/vllm-sim-pd/kustomization.yaml
Probably this is the right way to install

PR #258, opened 2025-11-18
update PD support to be compatible with nixlv2

### LINKED ISSUE #255: Make P-D simulation compatible with the latest version of nixl connector
Currently the sidecar for the D simulator in P-D scenario should be configured to use nixl connector, in this case it works with additional fields in the request/response (do_remote_decode, do_remote_prefill, remote_block_ids, remote_engine_id, remote_host, remote_port)

Current version of sidecar uses nixl2 connector which expects for the following fields:
- kv_transfer_params
  - do_remote_decode
  - do_remote_prefill
  - remote_engine_id
  - remote_block_ids
  - remote_host
  - remote_port

mayabar:
Fixes #255

### DIFF
--- a/pkg/openai-server-api/request.go
+++ b/pkg/openai-server-api/request.go
@@ -112,7 +112,8 @@ type KVTransferParams struct {
 	RemoteHost string `json:"remote_host"`
 	// RemotePort is a port of the remote server handling prefill
 	RemotePort int `json:"remote_port"`
-	TPSize     int `json:"tp_size" default:"1"`
+	// TPSize is the tensor parallelism size for KV cache transfer
+	TPSize int `json:"tp_size" default:"1"`
 }
 
 // StreamOptions defines streaming options for streaming requests
--- a/pkg/llm-d-inference-sim/test_utils.go
+++ b/pkg/llm-d-inference-sim/test_utils.go
@@ -19,6 +19,7 @@ import (
 	"bufio"
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


pkg/llm-d-inference-sim/simulator.go
```diff
@@ -576,14 +576,16 @@ func (s *VllmSimulator) createCompletionResponse(logprobs *int, isChatCompletion
 		time.Now().Unix(), modelName, usageData)
 
 	if doRemoteDecode {
+		baseResp.KVParams = &openaiserverapi.KVTransferParams{}
 		// add special fields related to the prefill pod special behavior
-		baseResp.DoRemoteDecode = true
-		baseResp.DoRemotePrefill = false
+		baseResp.KVParams.DoRemoteDecode = false
+		baseResp.KVParams.DoRemotePrefill = true
```
irar2, line 582: I don't understand why these two were flipped

pkg/openai-server-api/request.go
```diff
@@ -91,22 +91,28 @@ type baseCompletionRequest struct {
...
+	RemoteBlockIds []string `json:"remote_block_ids"`
 	// RemoteHost is a hostname or IP address of the remote server handling prefill
 	RemoteHost string `json:"remote_host"`
 	// RemotePort is a port of the remote server handling prefill
 	RemotePort int `json:"remote_port"`
-	// The number of tokens in the prompt that are in the local KV Cache
-	cachedPromptTokens int
-	// IgnoreEOS is a boolean value, true when the model should ignore end-of-sequence tokens
-	IgnoreEOS bool `json:"ignore_eos"`
+	TPSize     int `json:"tp_size" default:"1"`
```
irar2: Please add a comment describing TPSize

pkg/llm-d-inference-sim/simulator.go
mayabar: if doRemoteDecode is true - this means that prefill is executed now, the response should be configured to the next step (decode) for it doRemoteDecode shouls be false and doRemotePrefill should be true.
In past it worked because sidecar defined these fields and the current sidecar copies kv-transfer-params from the prefill response

pkg/openai-server-api/request.go
mayabar: added

Issue #249, opened 2025-11-02
Add support for N request parameter


mayabar:
Both text and chat completion requests have a parameter called `n` that defines how many choices should be returned in the response message.
Once this parameter is added, please ensure that the `request_max_num_generation_tokens` behaves as expected and add appropriate tests to verify it.

umakantv:
Hi @mayabar looks like the original PR is closed, I would like to take this up and address the comments you added on the original PR

mayabar:
@umakantv Welcome to the project ;)
Please pay attention that recently we added support for multiple inputs in /completions, this affects the N responses behaior.

umakantv:
Thanks @mayabar for the heads-up, I will review that.

umakantv:
Hello @mayabar please check this PR https://github.com/llm-d/llm-d-inference-sim/pull/537

This is my first task in this project and I have used some help from Claude in understanding this project and implementing this task, hope that is fine.

PR #262, opened 2025-11-29
pkg: add ability to provide number of choices sampled.

MadhavJivrajani:
To add the `N` parameter, each choice in the response is the same text content. 
This also helps simplify things a little bit in terms of testing and response construction.

Fixes https://github.com/llm-d/llm-d-inference-sim/issues/249

cc @mayabar


MadhavJivrajani:
Thanks! Let me rework it based on your comment.

PR #268, opened 2025-12-04
feat(metrics): add generation_tokens_total and prompt_tokens_total metrics

googs1025:


pkg/llm-d-inference-sim/metrics.go
```diff
@@ -338,14 +351,23 @@ func (s *VllmSimulator) setInitialPrometheusMetrics(cacheConfig *prometheus.Gaug
...
 		if s.config.FakeMetrics.TPOTBucketValues != nil {
 			s.initFakeHistogram(s.metrics.tpot, common.TPOTBucketsBoundaries, s.config.FakeMetrics.TPOTBucketValues)
-			s.initFakeHistogram(s.metrics.interTokenLatency, common.TPOTBucketsBoundaries, s.config.FakeMetrics.TPOTBucketValues)
 		}
 		buckets := build125Buckets(s.config.MaxModelLen)
 		if s.config.FakeMetrics.RequestPromptTokens != nil {
+			var promptTotal int64
 			s.initFakeHistogram(s.metrics.requestPromptTokens, buckets, s.config.FakeMetrics.RequestPromptTokens)
+			for _, v := range s.config.FakeMetrics.RequestPromptTokens {
+				promptTotal += int64(v)
```
mayabar: Values in the FakeMetrics.RequestPromptTokens means number of requests for each bucket. Summarizing all values in the array gives number of requests and not number of tokens.
Same for RequestGenerationTokens.
I would add an additional field for total number of tokens and support 2 cases:
- if user defined the total number of tokens - use it
- if total number of tokens is not defined - use middle value of each bucket * number of requests in this bucket for total number of tokens calculation.

googs1025: Oh, that's a bit embarrassing..... I misunderstood this part. You're right. I'll modify the implementation based on your suggestions.

googs1025: done

pkg/llm-d-inference-sim/metrics.go
```diff
@@ -764,3 +791,34 @@ func build125Buckets(maxValue int) []float64 {
...
+//
+// The last bucket is treated as [buckets[len(buckets)-1], +Inf), so its upper bound is approximated
+// as twice the lower bound for midpoint calculation.
+func estimateTokenTotal(counts []int, buckets []float64) int64 {
+	if len(counts) == 0 || len(buckets) == 0 {
+		return 0
+	}
+	var total int64
+	n := len(buckets)
+	for i, count := range counts {
```
mayabar: please check that buckets' length is >= counts' length

pkg/llm-d-inference-sim/metrics.go
```diff
@@ -764,3 +791,34 @@ func build125Buckets(maxValue int) []float64 {
...
+		if count == 0 {
+			continue
+		}
+		var lower, upper float64
+		lower = buckets[i]
```
mayabar: Buckets array contains upper limit for all buckets, except the last bucket which is not represented in the buckets array. 

The first bucket is [0, buckets[0]]
Bucket i is (buckets[i-1], bucket[i]]
The last bucket is (buckets[len(buckets)-1], +Inf)

pkg/llm-d-inference-sim/metrics_test.go
```diff
@@ -710,12 +705,46 @@ var _ = Describe("Simulator metrics", Ordered, func() {
 			Expect(metrics).To(ContainSubstring(getFloatBucketMetricLine(testModel, generationTokensMetricName, math.Inf(1), expectedCount)))
 			Expect(metrics).To(ContainSubstring(getFloatBucketMetricLine(testModel, promptTokensMetricName, math.Inf(1), expectedCount)))
 			Expect(metrics).To(ContainSubstring(getFloatBucketMetricLine(testModel, paramMaxTokensMetricName, math.Inf(1), expectedCount)))
+			Expect(metrics).To(MatchRegexp(`vllm:generation_tokens_total{model_name="testmodel"} 310`))
```
mayabar: In continuation to the previous comment:

In this example buckets array is [1, 2, 5, 10, 20, 50, 100, 200, 500, 1000] 
Counts contains 3 values: 10, 20, 30, 
So value should be 0.5*10+1.5*20+3.5*30=140 instead of 1.5*10+3.5*20+7.5*30=310

pkg/llm-d-inference-sim/metrics_test.go
```diff
@@ -946,3 +975,75 @@ var _ = Describe("build125Buckets", Ordered, func() {
...
+				expected: 0,
+			},
+			{
+				name:     "single bucket, single request",
+				counts:   []int{1},
+				buckets:  []float64{10},
+				expected: 15, // bucket [10, +Inf) → upper = 20, mid = (10+20)/2 = 15
+			},
+			{
+				name:     "two buckets, exact midpoints",
```
mayabar: same as the comment above regarding the buckets limits

pkg/llm-d-inference-sim/metrics.go
googs1025: thanks. fixed this part. plz take a look

pkg/llm-d-inference-sim/metrics.go
```diff
@@ -764,3 +791,47 @@ func build125Buckets(maxValue int) []float64 {
...
+// as twice the lower bound for midpoint calculation.
+func estimateTokenTotal(counts []int, buckets []float64) int64 {
+	if len(counts) == 0 || len(buckets) == 0 {
+		return 0
+	}
+	var total int64
+	nBuckets := len(buckets)
+	nCounts := len(counts)
+
+	for i := 0; i <= nBuckets; i++ {
```
mayabar: Usually the counts array will be shorter than the buckets array, seems better to iterate till counts' length.

pkg/llm-d-inference-sim/metrics.go
```diff
@@ -764,3 +791,47 @@ func build125Buckets(maxValue int) []float64 {
...
+		count := 0
+		if i < nCounts {
+			count = counts[i]
+		}
```
mayabar: in else of this if we can start this loop since got the end of the counts array
consider 
if i>= nCounts {
break
}
.. use counts[i]

pkg/common/config.go
```diff
@@ -252,13 +252,23 @@ type Metrics struct {
 	TPOTBucketValues []int `yaml:"tpot-buckets-values" json:"tpot-buckets-values"`
 	// RequestPromptTokens RequestGenerationTokens RequestParamsMaxTokens Histogram fake-observation arrays for init.
 	// Each value will be passed to Observe() once at start-up.
+	// exactly once during initialization. Additionally:
```
mayabar: please fix the sentence

pkg/llm-d-inference-sim/metrics.go
```diff
@@ -764,3 +791,47 @@ func build125Buckets(maxValue int) []float64 {
...
+		if count == 0 {
+			continue
+		}
+		var lower, upper float64
+		switch {
+		case i == 0:
+			// First bucket: [0, buckets[0]]
+			lower = 0.0
```
mayabar: lower can be initialized before the loop + processing of the last bucket could be done after the loop - in this case no need for switch

mayabar (changes requested):
some additional comments

pkg/llm-d-inference-sim/metrics.go
googs1025: done

pkg/llm-d-inference-sim/metrics.go
googs1025: done

pkg/common/config.go
googs1025: done

mayabar (approved):

PR #275, opened 2025-12-09
fix: Use "cmpl-" prefix for /completions response IDs (#270)

### LINKED ISSUE #270: Generate correct response IDs prefixes
/chat/completions and /completions response ID should be created with different prefixes

/completions prefix should be changed to "cmpl-"

RohanDSkaria:
This PR adds a 'textComplIDPrefix' constant for text completion responses, updates 'createCompletionResponse()' and 'createUsageChunk()'  to select the appropriate prefix based on endpoint type.

Fixes #270

### DIFF
--- a/pkg/llm-d-inference-sim/simulator.go
+++ b/pkg/llm-d-inference-sim/simulator.go
@@ -43,8 +43,6 @@ import (
 )
 
 const (
-	chatComplIDPrefix         = "chatcmpl-"
-	textComplIDPrefix         = "cmpl-"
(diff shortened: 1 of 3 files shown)


irar2:
It would probably look better to hide the choice of the prefix inside CreateChatCompletionResponse and CreateTextCompletionResponse functions (by concatenation of the correct prefix with the request id). And remove id argument from CreateBaseCompletionResponse.

Also please fix the sign-off.

PR #280, opened 2025-12-11
Support kv-events for /chat/completions requests too

mayabar:


pkg/llm-d-inference-sim/worker.go
```diff
@@ -89,12 +90,49 @@ func (s *VllmSimulator) processRequestAsync(reqCtx *openaiserverapi.CompletionRe
 			"metrics.lorasChan")
 	}
 
-	if s.config.EnableKVCache && !reqCtx.IsChatCompletion {
-		// kv cache is currently supported for /completion API only
-		if err := s.kvcacheHelper.OnRequestStart(req); err != nil {
+	if s.config.EnableKVCache {
```
irar2, line 92: Please consider moving to a separate function in helpers.go

PR #286, opened 2025-12-17
Latency calculator interface

irar2:


pkg/common/config.go
```diff
@@ -50,6 +50,8 @@ const (
 	RemoteDecodeFinishReason = "remote_decode"
 
 	podIPEnv = "POD_IP"
+
+	DefaultLatencyCalculator = "default"
```
elevran: nit: you can also use the empty string ("zero value") as the default instead of defining an explicit value. Nothing expliciitly set --> use the default value. This is backward comaptible (i.e., with configurations that lack the value completely).

pkg/llm-d-inference-sim/latencies.go
```diff
@@ -17,6 +17,15 @@ limitations under the License.
 // Package vllmsim implements the vLLM simulator.
 package llmdinferencesim
 
+type LatencyCalculator interface {
```
elevran, line 37: nit: Consider using return values of type `time.Duration` to make units explicit and not require conversion to msec.

pkg/llm-d-inference-sim/latencies.go
```diff
@@ -37,7 +46,7 @@ func (s *VllmSimulator) getPrefillTimePerToken() int {
 }
 
 // returns time to first token based on the current request's doRemotePrefill
-func (s *VllmSimulator) getWaitTimeToFirstToken(nPromptTokens int, nCachedPromptTokens int, doRemotePrefill bool) int {
+func (s *VllmSimulator) GetTimeToFirstToken(nPromptTokens int, nCachedPromptTokens int, doRemotePrefill bool) int {
```
elevran: nit: consider wrapping the existing time calculations in an object, so it is clear where a different LatencyCalculator should be instantiated and used. Currently, these are defined on the Simulator itself and it could be more instructive (and perhaps simplify testing?) if it is a distinct object.

pkg/common/config.go
irar2: done

pkg/llm-d-inference-sim/latencies.go
irar2: done

pkg/llm-d-inference-sim/latencies.go
irar2: done

Issue #259, opened 2025-11-20
Refactor requests and responses APIs


irar2:
Copied from #134, written by @frapschen 

The current implementation process of /v1/chat/completions and /v1/completions APIs:
```mermaid
flowchart TB
    rr@{ shape: circle, label: "Start" } --> ra["/v1/chat/completions"]
    rr@{ shape: circle, label: "Start" } --> rb["/v1/completions"]
    ra -->|s.HandleChatCompletions| A
    rb -->|s.HandleTextCompletions| A
    subgraph request-process
        A[handleCompletions] --> B{shouldInjectFailure}
        B -->|Yes| C[Return RandomFailure]
        B -->|No| E[parse raw request to a vllmReq]
        E --> F[validate vllmReq Request]
        F --> G[build CompletionReqCtx]
    end

    G -->|push to queue| H
    subgraph queue
        H@{ shape: das, label: "CompletionReqCtx channel" }
    end

    H -->|pop from queue| qb[reqCtx]
    subgraph reqProcessingWorker
        qb --> qc[mock metrics]
        qc --> qd[handle KVCache]
        qd --> qe[handel ToolCalls]
        qe --> qf{IsStream}
        qf -->|Yes| qg[return StreamingResponse]
        qg --> qj[build and write response chunk]
        qj --> ql[responseSentCallback]
        qf -->|No| qh[return Response]
        qh --> qk[build and write response]
        qk --> ql
        ql --> qm@{ shape: dbl-circ, label: "Stop" }
    end
```

Throughout the entire process, the `CompletionReqCtx` queue connects these two parts.

### request-process

In this flow, the main tasks performed are:
- Parsing the raw request body into a Go structize request
- Validating the structure of the request body
- Constructing a `CompletionReqCtx` and sending it to the queue

Currently, this process has high coupling, making it difficult to add new APIs. The following refactoring is required:

- Rename the `CompletionRequest` interface to `Request`, generalizing it to represent all potential future OpenAI requests. Add the following methods to this interface: 
  - `UnmarshalRequest`: Converts the raw request body into a specific request Go struct.
  - `ValidateRequest`: Validates the request Go struct.
  - `BuildRequestCtx`: Constructs a RequestCtx (this is a new interface, which will be discussed later) and sends it to the queue.
- Rename the `BaseCompletionRequest` struct to `BaseRequest`, generalizing it to represent the Go struct for all potential future OpenAI requests.  

[...]

PR #295, opened 2025-12-30
Process requests in requestContext

irar2:
#259 

This PR removes separate processor interface and moves request handling to requestContext

### DIFF
--- a/pkg/llm-d-inference-sim/requests.go
+++ b/pkg/llm-d-inference-sim/requests.go
@@ -80,48 +80,6 @@ func (b *baseRequestContext) done() {
 	b.wg.Done()
 }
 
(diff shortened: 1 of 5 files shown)


pkg/llm-d-inference-sim/requests.go
```diff
@@ -71,3 +79,103 @@ func (b *baseRequestContext) startProcessingTime() time.Time {
 func (b *baseRequestContext) done() {
 	b.wg.Done()
 }
+
+func (c *chatCompletionReqCtx) kvCacheOnRequestStart() *openaiserverapi.Error {
```
mayabar: consider moving chatCompletionReqCtx's function to chat_completion.go (same to text)

PR #296, opened 2025-12-30
feat: support cache threshold finish reason header to return cache_threshold finish reason

kyanokashi:
This PR adds support for returning a `cache_threshold` finish reason response. 

If the `X-Cache-Threshold` header is present in the request, the simulator will return the `cache_threshold` finish reason. 

WIP PR on the scheduler this is meant to support https://github.com/llm-d/llm-d-inference-scheduler/pull/509

### DIFF
--- a/pkg/llm-d-inference-sim/simulator_test.go
+++ b/pkg/llm-d-inference-sim/simulator_test.go
@@ -692,7 +692,7 @@ var _ = Describe("Simulator", func() {
 	})
 
 	Context("cache threshold finish reason header", func() {
-		It("Should return cache_threshold finish reason when header is set", func() {
+		testCacheThresholdFinishReasonHeader := func(setHeader bool, expectedFinishReason string) {
 			ctx := context.TODO()
 			client, err := startServer(ctx, common.ModeRandom)
 			Expect(err).NotTo(HaveOccurred())
@@ -706,7 +706,9 @@ var _ = Describe("Simulator", func() {
 			req, err := http.NewRequest("POST", "http://localhost/v1/chat/completions", strings.NewReader(reqBody))
 			Expect(err).NotTo(HaveOccurred())
 			req.Header.Set("Content-Type", "application/json")
-			req.Header.Set(cacheThresholdFinishReasonHeader, "true")
+			if setHeader {
+				req.Header.Set(cacheThresholdFinishReasonHeader, "true")
+			}
 
 			resp, err := client.Do(req)
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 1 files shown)


kfirwolfson:
> Hi @kyanokashi, The code looks good, but I'm uncertain about implementing this feature. As I understand, vLLM lacks this finish_reason—it's solely proposed in the [RFC](https://github.com/vllm-project/vllm/issues/24256).

Hi @irar2, the feature was presented and discussed in [KV-disaggregation](https://docs.google.com/document/d/1BSC5dGo_9npKkCBjpSGuN606IrkMUMh-bVs0Jg6eZQo/edit?tab=t.0) and [PD-disaggregation](https://docs.google.com/document/d/1ptMQSDRAYNq4xlE3USXrUROwF0c0rhAaIwNkcYnnOHI/edit?tab=t.0) SIGs, and got general acceptance. It's not just an RFC, there is also a vLLM [PR](https://github.com/vllm-project/vllm/pull/24520), which will hopefully get more traction when llm-d supports it.

> Even if pursued, checking actual KV cache usage against a threshold before triggering the finish reason would be preferable in my opinion.

Can you elaborate a bit - is this a check you expect to be added to the simulator?

kyanokashi:
Thank you for the pointers. Will work on getting this done promptly @mayabar

pkg/llm-d-inference-sim/simulator_test.go
```diff
@@ -689,4 +690,81 @@ var _ = Describe("Simulator", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			var chatResp map[string]interface{}
+			err = json.Unmarshal(body, &chatResp)
+			Expect(err).NotTo(HaveOccurred())
+
+			choices := chatResp["choices"].([]interface{})
+			Expect(choices).To(HaveLen(1))
+			firstChoice := choices[0].(map[string]interface{})
+			Expect(firstChoice["finish_reason"]).To(Equal("cache_threshold"))
```
mayabar: there is a constant for this string (CacheThresholdFinishReason)

pkg/llm-d-inference-sim/simulator_test.go
```diff
@@ -689,4 +690,81 @@ var _ = Describe("Simulator", func() {
 			Expect(string(body)).To(ContainSubstring("BadRequestError"))
 		})
 	})
+
+	Context("cache threshold finish reason header", func() {
+		It("Should return cache_threshold finish reason when header is set", func() {
```
mayabar: It("Should return cache_threshold finish reason when header is set", func() { 
and 
It("Should return normal finish reason when header is not set", func() {
are almost the same, please create a helper function for reuse

pkg/llm-d-inference-sim/streaming.go
```diff
@@ -138,7 +138,8 @@ func (s *VllmSimulator) sendTokenChunks(respCtx responseContext, ctx *fasthttp.R
 		var chunk openaiserverapi.CompletionRespChunk
 		var finishReasonToSend *string
 		if i == len(genTokens)-1 && (*respCtx.finishReason() == common.LengthFinishReason ||
-			*respCtx.finishReason() == common.ToolsFinishReason) {
+			*respCtx.finishReason() == common.ToolsFinishReason ||
+			*respCtx.finishReason() == common.CacheThresholdFinishReason) {
```
mayabar, line 142: just want to make sure that in streaming mode it's ok that cache threshold will never be returned as a separate chunk (like in the case of stop reason)

pkg/llm-d-inference-sim/simulator_test.go
```diff
@@ -689,4 +690,81 @@ var _ = Describe("Simulator", func() {
...
+			reqBody := `{
+            "messages": [{"role": "user", "content": "Hello"}],
+            "model": "` + testModel + `",
+            "max_tokens": 5
+        }`
+
+			req, err := http.NewRequest("POST", "http://localhost/v1/chat/completions", strings.NewReader(reqBody))
+			Expect(err).NotTo(HaveOccurred())
+			req.Header.Set("Content-Type", "application/json")
+			req.Header.Set("X-Cache-Threshold-Finish-Reason", "true")
```
mayabar: please use the constant we have for this string

mayabar (changes requested):
Hi @kyanokashi 
I added some comments, 
in addition, currently even if the `X-Cache-Threshold-Finish-Reason` is defined as true the simulator will wait the calculated "processing time", while in vllm the request will return immediately.

pkg/llm-d-inference-sim/streaming.go
kyanokashi: @kfirwolfson

PR #301, opened 2026-01-08
feat: cache hit threshold handling

kyanokashi:
This PR introduces support enabled by the kv-cache helper for handling request and global cache-hit-threshold parameters.

### DIFF
--- a/pkg/llm-d-inference-sim/request.go
+++ b/pkg/llm-d-inference-sim/request.go
@@ -45,7 +45,7 @@ type requestContext interface {
 	httpRequestCtx() *fasthttp.RequestCtx
 	done()
 	startProcessingTime() time.Time
-	kvCacheOnRequestStart() (float64, *openaiserverapi.Error)
+	kvCacheOnRequestStart() (hitRate float64, serverError *openaiserverapi.Error)
 	kvCacheOnRequestEnd()
 	createToolCalls() ([]openaiserverapi.ToolCall, int, string, error)
 	handleRequest() (responseContext, string, *openaiserverapi.Error)


pkg/common/config.go
```diff
@@ -781,6 +787,11 @@ func ParseCommandParamsAndLoadConfig() (*Configuration, error) {
 
 	f.BoolVar(&config.EnableKVCache, "enable-kvcache", config.EnableKVCache, "Defines if KV cache feature is enabled")
 	f.IntVar(&config.KVCacheSize, "kv-cache-size", config.KVCacheSize, "Maximum number of token blocks in kv cache")
+	var globalCacheHitThreshold float64 = -1 // Use -1 as sentinel for "not set"
+	if config.GlobalCacheHitThreshold != nil {
+		globalCacheHitThreshold = *config.GlobalCacheHitThreshold
+	}
+	f.Float64Var(&globalCacheHitThreshold, "global-cache-hit-threshold", -1, "Default cache hit threshold (0-1) for all requests. If a request specifies cache_hit_threshold, it takes precedence")
```
mayabar: setting default value as 0 is enough, no need to add an additional variable with value -1

pkg/common/config.go
```diff
@@ -869,6 +880,11 @@ func ParseCommandParamsAndLoadConfig() (*Configuration, error) {
 		}
 	}
 
+	// Set global cache hit threshold if provided via command line
+	if globalCacheHitThreshold >= 0 && globalCacheHitThreshold <= 1 {
+		config.GlobalCacheHitThreshold = &globalCacheHitThreshold
+	}
+
```
mayabar: see the comment above

pkg/common/config.go
```diff
@@ -869,6 +880,11 @@ func ParseCommandParamsAndLoadConfig() (*Configuration, error) {
...
 	if err := config.validate(); err != nil {
```
mayabar, line 883: please add to config.validation function validation that provided global cache threshold is in the required range [0, 1]

pkg/kv-cache/block_cache.go
```diff
@@ -287,6 +287,29 @@ func (bc *blockCache) getBlockInfo(blockHash uint64) (int, bool) {
...
+	count := 0
+	for _, blockHash := range blocks {
+		// Check if block is in used blocks (currently in use by running requests)
+		if _, exists := bc.usedBlocks[blockHash]; exists {
+			count++
+		} else if _, exists := bc.unusedBlocks[blockHash]; exists {
+			// Check if block is in unused blocks (was used in past)
+			count++
+		}
+	}
```
mayabar, line 311: A block in the middle of the sequence can be evicted — which count should this function return?
1. The length of the continuous (hole-free) sequence of blocks present in the cache
2. The total number of blocks currently in the cache, even if some are missing.
Could you please check how vLLM counts blocks to ensure consistency?

pkg/kv-cache/kv_cache.go
```diff
@@ -94,9 +107,40 @@ func (h *KVCacheHelper) OnRequestStart(vllmReq openaiserverapi.Request) error {
 		blockHashes[i] = key.ChunkHash
 	}
 
+	return blockHashes, nil
+}
+
+// GetCacheHitInfo returns cache hit information for a request
+func (h *KVCacheHelper) GetCacheHitInfo(vllmReq openaiserverapi.Request) (CacheHitInfo, error) {
```
mayabar: this function involves tokenization, and it invoked twice, maybe it's possible to return the cache hit from the OnRequestStart to avoid multiple tokenizations.

pkg/llm-d-inference-sim/request.go
```diff
@@ -131,8 +161,43 @@ func (reqCtx *baseRequestContext) handleRequest() (responseContext, string, *ope
...
+
+		if cacheHitThreshold != nil {
+			cacheHitInfo, err := reqCtx.kvCacheGetCacheHitInfo()
+			if err != nil {
+				return false, err
+			}
+
+			hitRate := cacheHitInfo.HitRate()
+			// If hit rate is below threshold, return cache_threshold finish reason
+			if hitRate < *cacheHitThreshold {
```
mayabar, line 182: do we want for cacheHitThreshold 0 behave like threshold is not defined?

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -375,32 +376,37 @@ func (s *VllmSimulator) responseSentCallback(reqCtx requestContext, model string
...
+			time.Sleep(perTokenLatency)
 
-		// report tpot in seconds
-		common.WriteToChannel(s.context.metrics.tpotChan, perTokenLatency.Seconds(), s.context.logger, "metrics.tpotChan")
+			// report tpot in seconds
+			common.WriteToChannel(s.context.metrics.tpotChan, perTokenLatency.Seconds(), s.context.logger, "metrics.tpotChan")
+		}
+		common.WriteToChannel(s.context.metrics.reqDecodeTimeChan, time.Since(startDecode).Seconds(), s.context.logger, "metrics.reqDecodeTimeChan")
 	}
-	common.WriteToChannel(s.context.metrics.reqDecodeTimeChan, time.Since(startDecode).Seconds(), s.context.logger, "metrics.reqDecodeTimeChan")
```
mayabar, line 404: Please check vllm's implementation regarding the processing time metrics reposts, what should be reported for request that were returned without processing? 
Should we skip them or report some short time?

pkg/llm-d-inference-sim/streaming.go
```diff
@@ -97,22 +104,28 @@ func (s *VllmSimulator) sendStreamingResponse(reqCtx requestContext, respCtx res
...
-	params := TTFTParams{
-		PromptTokens:       respCtx.usageData().PromptTokens,
-		CachedPromptTokens: respCtx.numberCachedPromptTokens(),
-		DoRemotePrefill:    respCtx.doRemotePrefill(),
-		RunningReqs:        s.context.metrics.nRunningReqs,
+
+	// Skip delays if finish reason is cache_threshold (immediate return)
+	isCacheThresholdFinishReason := respCtx.finishReason() != nil && *respCtx.finishReason() == common.CacheThresholdFinishReason
+
+	if !isCacheThresholdFinishReason {
```
mayabar, line 111: same as in non-streaming request regarding metrics

pkg/kv-cache/kv_cache.go
```diff
@@ -71,18 +71,31 @@ func (h *KVCacheHelper) Activate() {
 	h.blockCache.activate()
 }
 
-func (h *KVCacheHelper) OnRequestStart(vllmReq openaiserverapi.Request) error {
-	h.logger.V(logging.TRACE).Info("KV cache - process request")
+// CacheHitInfo contains information about cache hits for a request
```
mayabar: Do we really need `CacheHitInfo` struct? It seems like it’s only used to return the hit rate. Maybe we could just calculate the rate directly and use a float instead?

pkg/kv-cache/kv_cache.go
kyanokashi: That's what I did in the beginning 😅

pkg/llm-d-inference-sim/request.go
kyanokashi: Yes, that would be the only way to override a global cache hit threshold if defined

pkg/kv-cache/block_cache.go
kfirwolfson: Today vLLM supports only prefix reuse of KV cache. So only the first consecutive tokens will be considered as a hit.
We are [working on](https://github.com/vllm-project/vllm/issues/25950) changing this to allow smart recalculation of missing token cache in the middle of the prompt, but for now we should only count the consecutive prefix.

pkg/llm-d-inference-sim/simulator.go
kyanokashi: Correct me if I'm wrong but no token was generated and prefill or decode wasn't done so those wouldn't be valid metrics. @kfirwolfson are there other metrics you believe need to be reported?

kfirwolfson: Inside vLLM there is a metric counting total requests per `finish_reason`. I'm not sure it's implemented in the simulator, but if so, we can make sure it's also up to date with the `cache_threshold` reason.

Example:
```
$ curl http://0.0.0.0:8000/metrics | grep finished_reason
vllm:request_success_total{engine="0",finished_reason="stop",model_name="Llama-3.1-8B-Instruct"} 0.0
vllm:request_success_total{engine="0",finished_reason="length",model_name="Llama-3.1-8B-Instruct"} 1.0
vllm:request_success_total{engine="0",finished_reason="abort",model_name="Llama-3.1-8B-Instruct"} 0.0
vllm:request_success_total{engine="0",finished_reason="error",model_name="Llama-3.1-8B-Instruct"} 0.0
vllm:request_success_total{engine="0",finished_reason="cache_threshold",model_name="Llama-3.1-8B-Instruct"} 1.0
```

kfirwolfson: regarding the time metrics - good catch @mayabar!
I just tested on vLLM and it seems requests with finish_reason=cache_threshold are counted as an extremely low TTFT. I see it as a bug and will fix it in the vLLM PR. For that I'll have to check how requests ending with "aborted" or "error" finish_reason are counted and work according to that.

pkg/kv-cache/kv_cache.go
mayabar: If `CacheHitInfo` is not in use any more it can be deleted

pkg/llm-d-inference-sim/chat_completion.go
```diff
@@ -89,9 +89,9 @@ func (c *chatCompletionReqCtx) request() request {
 	return c.req
 }
 
-func (c *chatCompletionReqCtx) kvCacheOnRequestStart() *openaiserverapi.Error {
+func (c *chatCompletionReqCtx) kvCacheOnRequestStart() (float64, *openaiserverapi.Error) {
```
mayabar: please add a comment explaining the float64 meaning

mayabar (commented):
Hi @kyanokashi 
The PR look good to me, added some small comments

pkg/llm-d-inference-sim/chat_completion.go
kyanokashi: Instead I returned a named parameter `hitRate float64`

mayabar:
\lgtm

@kyanokashi you missed signoff in one commit

mayabar:
@kyanokashi please take a look under the title "Some checks were not successful", in DCO part
image
In addition, one test failed.- 
image

PR #330, opened 2026-02-12
Generation request and response types

irar2:


pkg/llm-d-inference-sim/generation.go
```diff
@@ -0,0 +1,136 @@
...
+		baseRequestContext: newBaseRequestContext(simCtx, respSender, wg),
+		req:                g,
+	}
+	// wire generationReqCtx into embedded requestContext interface
+	reqCtx.requestContext = reqCtx
+	return reqCtx
+}
+
+func (g *generationRequest) asString() string {
+	return "chat completion request (req id " + g.RequestID + ")"
```
mayabar: shouldn't be "generation" instead of "chat completion"?

PR #358, opened 2026-02-23
Add vllm:prefix_cache_hits and vllm:prefix_cache_queries counters

### LINKED ISSUE #356: Add vllm:prefix_cache_hits and vllm:prefix_cache_queries Prometheus counters
## What

Add the `vllm:prefix_cache_hits` and `vllm:prefix_cache_queries` Prometheus counters to match vLLM's prefix cache observability surface.

Both are token-granularity counters — `queries` increments by total prompt tokens on each request, `hits` increments by the number of tokens found already cached. This matches vLLM v1's semantics in `kv_cache_manager.py` where `prefix_cache_stats.record(num_tokens=request.num_tokens, num_hits=num_new_computed_tokens)`.

## Why

The simulator currently tracks KV cache *utilization* (`vllm:kv_cache_usage_perc`) but has no metric for cache *effectiveness*. When benchmarking prefix-cache-aware scorer strategies (e.g., `precise-prefix-cache-scorer` vs `prefix-cache-scorer`), there's no way to measure whether routing decisions actually result in higher cache reuse without scraping these counters.

Both counters are needed — `hits` alone is uninterpretable without the `queries` denominator. Together they give `rate(vllm:prefix_cache_hits[5m]) / rate(vllm:prefix_cache_queries[5m])` for a rolling hit rate.

## Implementation notes

The data is already computed in `pkg/kv-cache/kv_cache.go:OnRequestStart()`:
- `len(tokens)` → maps to `queries` increment
- `nBlocksAlreadyInCache * blockSize` → maps to `hits` increment

Wiring follows the existing pattern: add a channel + async updater goroutine in `metrics.go`, same as `kvCacheUsageChan` → `kvCacheUsageUpdater()`.

Only applies when `--enable-kvcache` is set — counters stay at zero otherwise (matching vLLM behavior when prefix caching is disabled).

## Ref

- Discussed in #347 (comment by @mayabar)
- vLLM source: `vllm/v1/metrics/loggers.py:509-516`, `vllm/v1/metrics/stats.py:115-142`

InfraWhisperer:
## Summary
- Add `vllm:prefix_cache_hits` and `vllm:prefix_cache_queries` Prometheus counters matching vLLM v1 token-level semantics
- `queries` increments by total prompt tokens per request, `hits` by cached tokens — enables `rate(hits) / rate(queries)` for cache effectiveness measurement
- Follows existing channel + async updater goroutine pattern (`kvCacheUsageChan` → `kvCacheUsageUpdater`)

Closes #356

## Test plan
- [x] Verify `go build ./...` passes
- [ ] Run simulator with `--enable-kvcache` and confirm both counters appear on `/metrics`
- [ ] Send repeated prompts with shared prefixes, verify `prefix_cache_hits` increments on subsequent requests
- [ ] Confirm counters stay at zero when `--enable-kvcache` is not set
- [ ] Validate `rate(vllm:prefix_cache_hits[5m]) / rate(vllm:prefix_cache_queries[5m])` produces expected hit rate in Prometheus

### DIFF
--- a/pkg/common/config.go
+++ b/pkg/common/config.go
@@ -316,6 +316,11 @@ type Metrics struct {
 	ReqPrefillTimeBucketValues []int `yaml:"prefill-time-buckets-values" json:"prefill-time-buckets-values"`
 	// ReqDecodeTimeBucketValues is an array of values for request decode time buckets.
 	ReqDecodeTimeBucketValues []int `yaml:"decode-time-buckets-values" json:"decode-time-buckets-values"`
+
+	// PrefixCacheHits is the initial value for the prefix cache hits counter (in tokens)
+	PrefixCacheHits *int64 `yaml:"prefix-cache-hits" json:"prefix-cache-hits,omitempty"`
+	// PrefixCacheQueries is the initial value for the prefix cache queries counter (in tokens)
+	PrefixCacheQueries *int64 `yaml:"prefix-cache-queries" json:"prefix-cache-queries,omitempty"`
 }
 
 type LorasMetrics struct {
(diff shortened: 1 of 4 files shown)


InfraWhisperer:
> @InfraWhisperer thanks for you PR Looks good, some general comments:
> 
> * please add support for fake metrics: check new values validity on initialization, e.g. check that values are non-negative, initialize prometheus values from the fake values given in the configuration, ...
> * tests for the new metrics are missing: need to test calculation in both scenarios, real and fake metrics

Hi @mayabar I have added tests as per your comments.

PR #396, opened 2026-03-18
Chat completion with kvcache

### LINKED ISSUE #180: Add support for KV Cache for /chat/completions API
Relevant for:

- KV events
- Prefill time calculations
- Prometheus metrics

mayabar:
Fixes #180


pkg/llm-d-inference-sim/chat_completion.go
```diff
@@ -104,7 +104,15 @@ func (c *chatCompletionReqCtx) encode() ([]uint32, []string, error) {
 }
 
 func (c *chatCompletionReqCtx) kvCacheOnRequestStart() (hitRate float64, oaiServerError *openaiserverapi.Error) {
-	// kv cache is currently supported for /completion API only
+	if c.sim.Config.EnableKVCache {
```
irar2: Maybe we can move the implementation to the base class now?

PR #402, opened 2026-03-21
Refactor configuration: To support pod name, namespace and dev mode in configuration parameters

### LINKED ISSUE #383: Read environment variables into configuration
There are some env variables that are not part of the configuration:
isInDevMode:    os.Getenv("VLLM_SERVER_DEV_MODE") == "1",
namespace:     os.Getenv(podNsEnv),
pod:           os.Getenv(podNameEnv),

Mrudhulraj:
This change resolves #383 : Eliminate env variables dependency and reads values from the config.yaml file parameter

- Added `PodName`, `PodNameSpace` and `VllmDevMode` fields to the Configuration struct
- Enabled reading these values from config file
- Added the fields in manifests/config.yaml file as a sample

Config Logs:
```I0321 01:32:21.794801   99636 config.go:811] "Configuration:" =<
        {
          "IP": "",
          "block-size": 16,
          "data-parallel-rank": -1,
          "data-parallel-size": 1,
          "dataset-in-memory": false,
          "dataset-path": "",
          "dataset-table-name": "llmd",
          "dataset-url": "",
          "default-embedding-dimensions": 384,
          "enable-kvcache": false,
          "enable-request-id-headers": false,
          "enable-sleep-mode": false,
          "event-batch-size": 16,
          "failure-injection-rate": 0,
          "failure-types": null,
          "fake-metrics": null,
          "fake-metrics-refresh-interval": 100000000,
          "global-cache-hit-threshold": 0,
          "hash-seed": "",
          "inter-token-latency": 1000000000,
          "inter-token-latency-std-dev": 0,
          "kv-cache-size": 1024,
          "kv-cache-transfer-latency": 100000000,
          "kv-cache-transfer-latency-std-dev": 0,
          "kv-cache-transfer-time-per-token": 0,
          "kv-cache-transfer-time-std-dev": 0,
          "latency-calculator": "",
          "lora-modules": [
            {
              "base_model_name": "",
              "name": "lora1",
              "path": "/path/to/lora1"
            },
            {
              "base_model_name": "",
              "name": "lora2",
              "path": "/path/to/lora2"
            }
          ],
          "max-cpu-loras": 5,
          "max-loras": 2,
          "max-model-len": 1024,
          "max-num-seqs": 5,
          "max-tool-call-array-param-length": 5,
          "max-tool-call-integer-param": 100,
          "max-tool-call-number-param": 100,
          "max-waiting-queue-length": 1000,
          "min-tool-call-array-param-length": 1,
          "min-tool-call-integer-param": 0,
          "min-tool-call-number-param": 0,
          "mode": "random",
          "model": "test",
          "object-tool-call-not-required-field-probability": 50,
          "pod-name": "POD_NAME_TEST",  --------------------------------> Read from Config  file
          "pod-namespace": "POD_NAMESPACE_TEST", --------------------------------> Read from Config File
          "port": 5000,
          "prefill-overhead": 0,
          "prefill-time-per-token": 0,
          "prefill-time-std-dev": 0,
          "seed": 100100100,
          "self-signed-certs": false,
          "served-model-name": [
            "model1",
            "model2"
          ],
          "ssl-certfile": "",
          "ssl-keyfile": "",
          "time-factor-under-load": 1,
          "time-to-first-token": 2000000000,
          "time-to-first-token-std-dev": 0,
          "tool-call-not-required-param-probability": 50,
          "uds-socket-path": "/tmp/tokenizer/tokenizer-uds.socket",
          "vllm-dev-mode": true,
          "zmq-endpoint": "tcp://localhost:5557",
          "zmq-max-connect-attempts": 0
        }
```


pkg/llm-d-inference-sim/simulator.go
```diff
@@ -60,8 +54,7 @@ type VllmSimulator struct {
 
 	// indication whether the simulator is sleeping
 	IsSleeping bool
-	// indication whether the simulator is in development mode, set by environment
-	// variable VLLM_SERVER_DEV_MODE
+	// indication whether the simulator is in development mode
 	IsInDevMode bool
```
irar2: We can now remove this field completely since we have in the configuration.

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -79,33 +72,29 @@ type VllmSimulator struct {
...
 		Context: SimContext{
 			logger: logger,
 			loras: &lorasUsageInfo{
 				loadedLoras: make(map[string]int),
 			},
 			kvcacheHelper: nil, // kvcache helper will be created only if required after reading configuration
-			Namespace:     os.Getenv(PodNsEnv),
-			Pod:           os.Getenv(PodNameEnv),
+			Namespace:     cfg.PodNameSpace,
+			Pod:           cfg.PodName,
```
irar2: No need for these fields

pkg/common/config.go
```diff
@@ -565,6 +595,212 @@ func (c *Configuration) Copy() (*Configuration, error) {
...
+	f.IntVar(&config.MaxNumSeqs, "max-num-seqs", config.MaxNumSeqs, "Maximum number of inference requests that could be processed at the same time")
+	f.IntVar(&config.MaxWaitingQueueLength, "max-waiting-queue-length", config.MaxWaitingQueueLength, "Maximum length of inference requests waiting queue")
+	f.IntVar(&config.MaxLoras, "max-loras", config.MaxLoras, "Maximum number of LoRAs in a single batch")
+	f.IntVar(&config.MaxCPULoras, "max-cpu-loras", config.MaxCPULoras, "Maximum number of LoRAs to store in CPU memory")
+	f.IntVar(&config.MaxModelLen, "max-model-len", config.MaxModelLen, "Model's context window, maximum number of tokens in a single request including input and output")
+
+	f.StringVar(&config.PodName, "pod-name", config.PodName, "Specifies the name of the pod running the simulator instance")
+	f.StringVar(&config.PodNameSpace, "pod-namespace", config.PodNameSpace, "Specifies the Kubernetes namespace in which the simulator pod is running")
+	f.BoolVar(&config.VllmDevMode, "vllm-dev-mode", config.VllmDevMode, "Enables development mode for the vLLM simulator")
+
```
irar2: Please remove these, we only want to read these fields from the environment

pkg/common/config.go
```diff
@@ -565,6 +595,212 @@ func (c *Configuration) Copy() (*Configuration, error) {
...
+	if podNameSpace == "" {
+		config.PodNameSpace = podNsEnv
+	} else {
+		config.PodNameSpace = podNameSpace
+	}
+
+	if os.Getenv("VLLM_SERVER_DEV_MODE") == "1" {
+		config.VllmDevMode = true
+	}
+
```
irar2: Please change to simply:
config.PodName = os.Getenv(PodNsEnv)
config.PodNameSpace = os.Getenv(podNsEnv)
config.VllmDevMode = os.Getenv("VLLM_SERVER_DEV_MODE") == "1"

Or maybe add a constant for "VLLM_SERVER_DEV_MODE" as well

pkg/llm-d-inference-sim/simulator.go
Mrudhulraj: Should I remove the Namespace and Pod field from the SimContext struct too? since we have it in the config.
I will refer the set config params in the rest of the code.

irar2: yes please

pkg/common/config.go
irar2: Why did you remove constants for "POD_NAME" and "POD_NAMESPACE"?

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -79,33 +68,26 @@ type VllmSimulator struct {
 }
 
 // New creates a new VllmSimulator instance with the given logger
-func New(logger logr.Logger) (*VllmSimulator, error) {
+func New(ctx context.Context, cfg *common.Configuration, logger logr.Logger) (*VllmSimulator, error) {
```
irar2: no need for these new arguments

pkg/common/config.go
```diff
@@ -100,6 +106,19 @@ type Configuration struct {
...
+	PodNameSpace string `yaml:"pod-namespace" json:"pod-namespace"`
+	// PodName specifies the name of the pod running the simulator instance.
+	// Used for identification in Kubernetes environments.
+	// Set by env variable POD_NAME
+	PodName string `yaml:"pod-name" json:"pod-name"`
+	// VllmDevMode enables development mode for the vLLM simulator
+	// Allowing for additional debugging features during local development and testing.
+	// Set by env variable VLLM_SERVER_DEV_MODE
+	VllmDevMode bool `yaml:"vllm-dev-mode" json:"vllm-dev-mode"`
+
```
irar2, line 115: No need in yaml and json tags here

pkg/common/parser.go
```diff
@@ -28,7 +28,10 @@ import (
 )
 
 const (
-	dummy = " "
+	dummy                = " "
+	vllmServerDevModeEnv = "VLLM_SERVER_DEV_MODE"
+	podNameEnv           = "POD_NAME"
+	podNsEnv             = "POD_NAMESPACE"
```
irar2: Please make pod... public and use them in the tests

pkg/tests/simulator_test.go
```diff
@@ -43,6 +42,10 @@ import (
 
 const prompt1 = "What is the weather like in New York today?"
 const prompt2 = "I hear it's very cold."
+const (
+	podNameEnv = "POD_NAME"
+	podNsEnv   = "POD_NAMESPACE"
+)
```
irar2: Please remove and use the constants defined in common

pkg/common/parser.go
Mrudhulraj: should I make the vllmServerDevMode too public?

PR #406, opened 2026-03-23
Remove CGO dependency by migrating to pure-Go ZMQ+change in ci_pr_checks

mayabar:


pkg/common/publisher.go
```diff
@@ -20,54 +20,51 @@ import (
...
-		// If not the last attempt, wait before retrying
-		if i < retries {
-			time.Sleep(1 * time.Second)
+	// 2. Push Dial into a background goroutine
+	go func() {
+		// wait until the listener is ready
+		err := socket.Dial(endpoint)
+		if err != nil {
+			// This only triggers if the address format is invalid or ctx is canceled
+			log.FromContext(ctx).Error(err, "Simulator ZMQ dialer exited", "endpoint", endpoint)
```
irar2: Log on success

mayabar: done

.github/workflows/ci-pr-checks.yaml
```diff
@@ -1,42 +1,72 @@
...
   lint-and-test:
+    needs: check-changes
+    if: ${{ needs.check-changes.outputs.src == 'true' }}    
     runs-on: ubuntu-latest
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+
```
shmuelk: Please revert this change

PR #427, opened 2026-04-13
Use kv-cache anduds-tokenizer version 0.7.1

mayabar:


pkg/tests/http_server_test.go
```diff
@@ -82,8 +82,8 @@ var _ = Describe("Server", func() {
 			var tokenizeResp vllmapi.TokenizeResponse
 			err = json.Unmarshal(body, &tokenizeResp)
 			Expect(err).NotTo(HaveOccurred())
-			Expect(tokenizeResp.Count).To(Equal(22))
-			Expect(tokenizeResp.Tokens).To(HaveLen(22))
+			Expect(tokenizeResp.Count).To(Equal(19))
+			Expect(tokenizeResp.Tokens).To(HaveLen(19))
```
shmuelk, line 86: In a followup PR please don't use the default of "include special characters" in the tokenization call.

mayabar: added in this PR

PR #435, opened 2026-04-19
Graceful shutdown: drain in-flight requests before shutting down

irar2:
- Stop accepting new requests immediately on shutdown signal (HTTP 503 / gRPC Unavailable)
- Wait up to 30 seconds for all in-flight requests to complete before tearing down the transport
- Internal goroutines (workers, metrics, kvcache, tokenizer) now run on a drainCtx derived from context.Background() so they stay alive through the drain period after the parent context is cancelled

### DIFF
--- a/pkg/communication/http.go
+++ b/pkg/communication/http.go
@@ -18,9 +18,7 @@ package communication
 
 import (
 	"bufio"
-	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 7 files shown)


pkg/communication/http.go
```diff
@@ -90,52 +89,31 @@ func (c *Communication) StartHTTPServer(ctx context.Context, listener net.Listen
...
-		return nil
+	return server, errCh, nil
+}
 
-	case err := <-serverErr:
-		if err != nil {
-			c.logger.Error(err, "server failed")
-		}
+// StartHTTPServer starts the HTTP server on the given listener and blocks until it exits.
+// Intended for use in tests with a custom listener. Shutdown is driven by closing the listener.
```
mayabar: consider moving this function to test_utils.go file (used by tests only)

pkg/llm-d-inference-sim/simulator.go
```diff
@@ -421,6 +448,12 @@ func (s *VllmSimulator) ResponseSentCallback(reqCtx requestContext) {
 	reqCtx.kvCacheOnRequestEnd()
 }
 
+// OpenRequests returns the number of requests currently in the system
+// (waiting in queue + being processed by workers).
+func (s *VllmSimulator) OpenRequests() int64 {
+	return s.openRequests.Load()
```
mayabar: if it possible to check length of the waiting queue + number of working workers instead of managing an additional counter

Issue #277, opened 2025-12-09
Confirm KV cache functions correctly when using LoRAs


mayabar:
- Check real vLLM's behavior when running with multiple loras. 
- Scenario to use: send same prompt to two different loras, check if BockStored send twice.
- Implement the same behavior in the simulator.

mayabar:
Fixed by #441

PR #441, opened 2026-04-21
Extend kv-cache to manage prefixes per lora and support additional kv-event fields

### LINKED ISSUE #272: Fill all KV event's fields
Currently only BlockHashes is filled in BlockStored kv-event.
Fill all fields:
	BlockHashes     []any    // Changed from []uint64
	ParentBlockHash any      // Changed from *uint64
	TokenIds        []uint32
	BlockSize       int
	LoraID          *int    `msgpack:",omitempty"`
	Medium          *string `msgpack:",omitempty"`

mayabar:
## Summary
- LoRA-aware block cache: Block keys now include model name alongside the hash, so blocks belonging to different LoRAs/base model are distinguished. Eviction prioritizes blocks of unloaded models before evicting blocks of loaded ones.
- LoRA ID assignment: Each loaded LoRA is assigned a stable 1-based index (loraIDs slot). IDs are freed when a LoRA is unloaded and reassigned on load. The base model has no LoRA ID (0).
- LoRA fields in KV events: BlockStoredEvent now carries LoraName, LoraID, and ParentHash fields, propagated through EventData and the msgpack publisher.

Fixes #272

### DIFF
--- a/pkg/kv-cache/block_cache.go
+++ b/pkg/kv-cache/block_cache.go
@@ -39,7 +39,7 @@ const (
 // contains sub-set of openai server api request fields that are relevant for the block cache
 type Request interface {
 	GetRequestID() string
-	GetModel() string
+	GetDisplayedModel() string
 	GetLoraName() *string
 	GetLoraID() *int
 }
@@ -180,7 +180,7 @@ func (bc *blockCache) startRequest(req Request, blockHashes []uint64, blockToken
 	// count number of new blocks + number of blocks that are in the unused blocks
 	// don't update the data until we are sure that it's ok
 	for i, blockHash := range blockHashes {
-		bKey := blockKey{hash: blockHash, modelName: req.GetModel()}
+		bKey := blockKey{hash: blockHash, modelName: req.GetDisplayedModel()}
 		if _, exists := bc.unusedBlocks[bKey]; exists {
 			blockToMoveToUsed = append(blockToMoveToUsed, bKey)
 		} else if _, exists := bc.usedBlocks[bKey]; !exists {
@@ -250,7 +250,7 @@ func (bc *blockCache) startRequest(req Request, blockHashes []uint64, blockToken
 	// store blockKeys and not only plain uint64
 	bc.requestToBlocks[req.GetRequestID()] = make([]blockKey, len(blockHashes))
 	for i, blockHash := range blockHashes {
-		bKey := blockKey{hash: blockHash, modelName: req.GetModel()}
+		bKey := blockKey{hash: blockHash, modelName: req.GetDisplayedModel()}
 		bc.requestToBlocks[req.GetRequestID()][i] = bKey
 	}
 
--- a/pkg/kv-cache/block_cache.go
+++ b/pkg/kv-cache/block_cache.go
@@ -124,7 +124,7 @@ func (bc *blockCache) activate() {
 
 // startRequest adds a request with its associated block hashes to the cache
 // and returns the number of blocks that were already in the cache
-func (bc *blockCache) startRequest(requestID string, blockHashes []uint64, blockTokens [][]uint32) (int, error) {
+func (bc *blockCache) startRequest(requestID string, loraName *string, blockHashes []uint64, blockTokens [][]uint32) (int, error) {
 	bc.mu.Lock()
 	defer bc.mu.Unlock()
 
@@ -217,7 +217,13 @@ func (bc *blockCache) startRequest(requestID string, blockHashes []uint64, block
 	}
 
 	if len(hashes) > 0 {
-		common.WriteToChannel(bc.eventChan, EventData{action: eventActionStore, hashes: hashes, tokens: tokens}, bc.logger)
+		common.WriteToChannel(bc.eventChan,
+			EventData{
+				action:   eventActionStore,
+				hashes:   hashes,
+				tokens:   tokens,
+				loraName: loraName,
+			}, bc.logger)
 	}
 
 	// store the request mapping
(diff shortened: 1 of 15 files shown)


pkg/kv-cache/block_cache.go
```diff
@@ -35,13 +35,28 @@ const (
...
 	mu              sync.RWMutex
-	requestToBlocks map[string][]uint64                // request id -> array of it blocks (block hashes)
-	usedBlocks      map[uint64]int                     // block hash -> reference count
-	unusedBlocks    map[uint64]time.Time               // block hash -> last usage timestamp
-	blockToTokens   map[uint64][]uint32                // block hash -> block tokens
+	requestToBlocks map[string][]blockKey              // request id -> array of it blocks (block hashes)
+	usedBlocks      map[blockKey]int                   // block hash -> reference count
+	unusedBlocks    map[blockKey]time.Time             // block hash -> last usage timestamp
+	blockToTokens   map[blockKey][]uint32              // block hash -> block tokens
+	loadedModels    map[string]bool                    // models currently loaded (base model + loaded loras)
```
irar2: Consider using map[string]{} here

pkg/kv-cache/block_cache.go
```diff
@@ -146,25 +163,26 @@ func (bc *blockCache) startRequest(requestID string, blockHashes []uint64, block
...
 
 	// first step - ensure that there is enough space for all blocks
 	// count number of new blocks + number of blocks that are in the unused blocks
 	// don't update the data until we are sure that it's ok
 	for i, blockHash := range blockHashes {
-		if _, exists := bc.unusedBlocks[blockHash]; exists {
-			blockToMoveToUsed = append(blockToMoveToUsed, blockHash)
-		} else if _, exists := bc.usedBlocks[blockHash]; !exists {
-			blocksToAdd = append(blocksToAdd, blockHash)
+		bKey := blockKey{hash: blockHash, modelName: vllmReq.GetModel()}
```
irar2: Do we need display model name here?

pkg/kv-cache/block_cache.go
```diff
@@ -183,46 +201,51 @@ func (bc *blockCache) startRequest(requestID string, blockHashes []uint64, block
...
+			}, bc.logger)
 	}
 
 	// store the request mapping
-	bc.requestToBlocks[requestID] = make([]uint64, len(blockHashes))
-	copy(bc.requestToBlocks[requestID], blockHashes)
+	// store blockKeys and not only plain uint64
+	bKeys := make([]blockKey, len(blockHashes))
+	for i, blockHash := range blockHashes {
+		bKey := blockKey{hash: blockHash, modelName: vllmReq.GetModel()}
```
irar2: Likewise

pkg/kv-cache/block_cache.go
```diff
@@ -183,46 +201,51 @@ func (bc *blockCache) startRequest(requestID string, blockHashes []uint64, block
...
+		bKeys[i] = bKey
+	}
+	bc.requestToBlocks[vllmReq.GetRequestID()] = make([]blockKey, len(blockHashes))
+	copy(bc.requestToBlocks[vllmReq.GetRequestID()], bKeys)
```
irar2: Maybe copy the keys directly in the loop?

pkg/kv-cache/kv_cache.go
```diff
@@ -125,7 +130,7 @@ func (h *KVCacheHelper) OnRequestStart(vllmReq openaiserverapi.Request) (float64
 	vllmReq.SetNumberOfCachedPromptTokens(cachedTokens)
 
 	totalBlocks := len(blockHashes)
-	cachedBlocks := h.blockCache.countCachedBlockPrefix(blockHashes)
+	cachedBlocks := h.blockCache.countCachedBlockPrefix(blockHashes, vllmReq.GetModel())
```
irar2: Display name here too

pkg/kv-cache/kv_cache_test.go
```diff
@@ -442,7 +467,12 @@ var _ = Describe("KV cache", Ordered, func() {
 							hashes, tokens := createRandomArray(testCase.minBlockLen, testCase.maxBlockLen,
 								testCase.maxHashValue, random)
 
-							_, err := blockCache.startRequest(reqID, hashes, tokens)
+							req := testRequest{
+								id:          reqID,
+								blockHashes: hashes,
+								tokens:      tokens,
+							}
+							_, err := blockCache.startRequest(&req, hashes, tokens)
```
irar2, line 477: Please add tests for different models/loras, eviction of unloaded loras, etc.

pkg/kv-cache/block_cache.go
shmuelk: I think Ira meant map[string]struct{}.

That gets the point accross that the key is important and the value is meaningless, unlike map[string]bool which can imply that the value is also important. A lesson I learned from others in llm-d.

mayabar: Done

pkg/kv-cache/block_cache.go
mayabar: Done

pkg/kv-cache/kv_cache_test.go
mayabar: Done

pkg/kv-cache/block_cache.go
mayabar: updated model name used in all places in the simulator

mayabar:
Fixes #277

PR #444, opened 2026-04-23
fix: return LoRA model path in root field of /v1/models (#443)

### LINKED ISSUE #443: Return LoRA Model Path in root for /v1/models
**Description**
The simulator currently does not return the LoRA model’s path in the root field of the /v1/models response. 

**Expected Behavior**
For LoRA models, the /v1/models endpoint should set root to the LoRA model path.

```{
  "object": "list",
  "data": [
    {
      "id": "lora1",
      "object": "model",
      "created": 1776688922,
      "owned_by": "vllm",
      "root": "path/to/lora1",
      "parent": "base_model_alias",
      "max_model_len": 32768
    }
  ]
}

### PR DESCRIPTION
**Pull request by mvanhorn**
Closes #443.

## Problem

LoRA entries in the \`/v1/models\` response used the adapter's name as both \`id\` and \`root\`. Real vLLM returns the LoRA's filesystem path as \`root\`, which the issue reports is what consumers expect.

## Root cause

\`SimContext.loraAdaptors\` is a \`sync.Map\` keyed by LoRA name. The path from \`LoraModule.Path\` (static config, \`context.go:82\`) and \`loadLoraRequest.LoraPath\` (dynamic load, \`lora.go:61\`) was being discarded at insertion — both call sites wrote an empty string as the value. With no path kept anywhere, \`CreateModelsResponse\` fell back to using the name for \`root\`.

## Fix

- Store the path as the map value at both insertion sites (config init and the load endpoint).
- Add a \`getLoraPath\` helper that reads the path back, falling back to the name when no path was supplied (preserves existing behavior for the dynamic-load-without-path case and for tests that store non-string sentinel values).
- Use \`getLoraPath(lora)\` for \`Root\` in \`CreateModelsResponse\`.

## After

\`\`\`json
{
  "id": "lora1",
  "object": "model",
  "created": 1776688922,
  "owned_by": "vllm",
  "root": "path/to/lora1",
  "parent": "base_model_alias",
  "max_model_len": 32768
}
\`\`\`

## Test

Added a spec to \`models_test.go\` that stores a path for one adapter and an empty string for another, and verifies \`root\` reflects the recorded path (falling back to name when none was set).

Ran:
- \`ginkgo --timeout=300s ./pkg/llm-d-inference-sim/ ./pkg/common/\` — 148/148 passing
- \`golangci-lint run ./pkg/llm-d-inference-sim/...\` — 0 issues

### DIFF
--- a/pkg/llm-d-inference-sim/context.go
+++ b/pkg/llm-d-inference-sim/context.go
@@ -79,7 +79,7 @@ func (s *SimContext) initialize(ctx context.Context) error {
 	}
 
 	for _, lora := range s.Config.LoraModules {
-		s.loraAdaptors.Store(lora.Name, "")
+		s.loraAdaptors.Store(lora.Name, lora.Path)
 	}
 	s.loras.maxLoras = s.Config.MaxLoras
 	s.loras.loraRemovable = common.Channel[int]{
@@ -224,7 +224,7 @@ func (s *SimContext) CreateModelsResponse() *vllmapi.ModelsResponse {
 			Object:      vllmapi.ObjectModel,
 			Created:     time.Now().Unix(),
 			OwnedBy:     "vllm",
-			Root:        lora,
+			Root:        s.getLoraPath(lora),
 			Parent:      &parent,
 			MaxModelLen: s.Config.MaxModelLen,
 		})
--- a/pkg/llm-d-inference-sim/lora.go
+++ b/pkg/llm-d-inference-sim/lora.go
@@ -49,6 +49,20 @@ func (s *SimContext) getLoras() []string {
 	return loras
 }
 
+// getLoraPath returns the filesystem path recorded for the given LoRA name.
+// Falls back to the name when no path was supplied (dynamic load without
+// lora_path, or the non-string sentinel values stored by some tests).
+func (s *SimContext) getLoraPath(name string) string {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by irar2** [comment]
@mvanhorn Thanks for the PR.

Please make sure to store the path when it’s set via the command line/config file in initialize() in context.go:
```
for _, lora := range s.Config.LoraModules {
		s.loraAdaptors.Store(lora.Name, "")
	}
```

Also, please check the 'root' field in the test at pkg/tests/lora_test.go.

mvanhorn:
Thanks for the review, @irar2!

**re: initialize() in context.go** — storing the path from config is already part of this PR at context.go:82: `s.loraAdaptors.Store(lora.Name, lora.Path)` (replacing the prior `Store(lora.Name, "")`). The snippet in your comment matches the pre-PR code; the committed diff (0803d1e) already makes that change at the insertion site you called out, and symmetrically in `LoadLoraAdaptor` so /load_lora_adapter also records the path.

**re: 'root' field test** — added in 8e9e4f7. Asserts `model.Root == expectedRoots[model.ID]` for lora1 (dynamic-load via /load_lora_adapter), lora3, and lora4 (config-loaded via --lora-modules) in the existing /models response loop.

Let me know if you had a different change in mind for context.go — happy to adjust if I'm reading the comment wrong.

PR #446, opened 2026-04-23
fix: support structured content blocks in echo mode for chat/completions

### LINKED ISSUE #438: Echo mode: support structured content blocks in chat/completions responses
## Problem
Echo mode returns empty responses for `/v1/chat/completions` requests when message content is structured (array of content blocks) rather than a plain string. The issue is in `tokenizedPromptForEcho()` in `chat_completion.go` which only reads `Content.Raw`.

## Proposed Solution
- Add a new ReadableText() method on Content that serializes all block types into a readable string with blocks separated by newlines. Do not change PlainText().
- Update tokenizedPromptForEcho() to use the new method.

## Serialization per block type

Block Type | Serialization
-- | --
text | The text value as-is
image_url | image: \<url\>

## Examples

Request: 
`"content": [{"type": "text", "text": "Describe this"}, {"type": "image_url", "image_url": {"url": "https://..."}}]`
Response text: 
```
Describe this
image: https://...
```
## Acceptance Criteria
- Echo mode returns serialized content of the last message when content is structured
- All block types (text and image_url) are represented
- Blocks separated by newlines
- Existing behavior for plain string content unchanged
- PlainText() not modified
- Unit tests cover structured content in echo mode

yizhaodev:
## Summary

Fixes #438

Echo mode returned empty responses for `/v1/chat/completions` when message content used structured blocks (array of `{type, text}` objects) instead of a plain string. The root cause was `tokenizedPromptForEcho()` reading `Content.Raw` directly, which is empty for structured content.

- Added `ReadableText()` method on `Content` that serializes all block types into a readable string (text as-is, image_url as `image: <url>`, blocks separated by newlines)
- Updated `tokenizedPromptForEcho()` to use `ReadableText()` instead of `Content.Raw`
- `PlainText()` is not modified

### DIFF
--- a/pkg/tests/simulator_test.go
+++ b/pkg/tests/simulator_test.go
@@ -358,6 +358,86 @@ var _ = Describe("Simulator", func() {
 		Entry(nil, common.MMModelName, common.ModeEcho, 10),
 	)
 
+	It("echo mode with structured content blocks", func() {
+		ctx := context.TODO()
+		args := []string{"cmd", "--model", common.TestModelName, "--mode", common.ModeEcho}
+		client, err := startServerWithArgs(ctx, args)
+		Expect(err).NotTo(HaveOccurred())
+
+		openaiclient := openai.NewClient(
+			option.WithBaseURL(baseURL),
+			option.WithHTTPClient(client),
+			option.WithMaxRetries(0))
+
+		params := openai.ChatCompletionNewParams{
+			Messages: []openai.ChatCompletionMessageParamUnion{
+				openai.UserMessage(
+					[]openai.ChatCompletionContentPartUnionParam{
+						openai.TextContentPart("Describe this"),
+						openai.ImageContentPart(openai.ChatCompletionContentPartImageImageURLParam{
+							URL: "https://example.com/img.png",
+						}),
+					},
+				),
+			},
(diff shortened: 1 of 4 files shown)


pkg/tests/simulator_test.go
```diff
@@ -358,6 +358,111 @@ var _ = Describe("Simulator", func() {
...
+		reqBody := `{
+			"messages": [{"role": "user", "content": [
+				{"type": "text", "text": "Describe this"},
+				{"type": "image_url", "image_url": {"url": "https://example.com/img.png"}}
+			]}],
+			"model": "` + common.TestModelName + `",
+			"stream": false
+		}`
+
+		resp, err := client.Post("http://localhost/v1/chat/completions", "application/json", strings.NewReader(reqBody))
```
mayabar: For regular requests (without special headers) we usually use openaiapi (github.com/openai/openai-go/v3)

mayabar (commented):
@yizhaodev Thanks for your PR. It looks good, just a small comment

pkg/tests/simulator_test.go
yizhaodev: thanks @mayabar , this is fixed now.

PR #447, opened 2026-04-24
SIM_MODEL override + optional HTTP request/response logging (log-http)

jland-redhat:
**Summary**
Implements reviewer feedback from [#447](https://github.com/llm-d/llm-d-inference-sim/pull/447): split from HTTP logging; env-driven overrides follow the usual precedence command-line → environment → YAML → defaults.

Behavior

**SIM_MODEL**: applied only when --model is not passed; overrides model from YAML/defaults. If --model is set, it always wins.
**PYTHONHASHSEED**: applied only when --hash-seed is not passed; overrides hash-seed from YAML/defaults. If --hash-seed is set, it always wins.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,10 +1,20 @@
 # Command line parameters
 The simulator can be configured using either command-line arguments or a YAML file. Parameter names are consistent across both methods.
 
+## Configuration precedence
+For a setting that can come from a YAML file, an environment variable, and command-line flags, the simulator resolves the value in this order (first wins):
+
+1. **Command-line flags** — for example `--model` or `--hash-seed`.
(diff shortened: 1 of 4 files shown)


irar2:
@jland-redhat Thanks a lot for the PR.

Could you please split this into two separate PRs, since it introduces two distinct features? That will make it easier to review and track.

Regarding the `SIM_MODEL` part: after looking into it, the commonly accepted precedence order is:
command-line flags (highest priority), then environment variables, then config files, and finally hardcoded defaults.

This ordering should address the issue you’re tackling. We also have an existing environment variable, `PYTHONHASHSEED`, that doesn’t follow this pattern - so it would be great if you could align its behavior as part of this change.

Please also add a short explanation of the precedence order to the `configuration.md`.

The HTTP logging part looks fine.

jland-redhat:
Fair comments will try to take care of this weekend

PR #457, opened 2026-04-28
Add support for /health/ready endpoint

### LINKED ISSUE #453: [Feature] Add /health/ready endpoint
vLLM is going to add /health/ready endpoint for GPU health verification

- https://github.com/vllm-project/vllm/issues/36960
- https://github.com/vllm-project/vllm/pull/36961

**Why is this needed**:
Simulator behavior can be the same, just to support the endpoint. This will help to unify the deployment yamls files.

mayabar:
In vLLM  /health/ready returns success status (code 200) when the engine is ready to process inference requests.
Simulate this by adding a new command line parameter `--startup-duration` which defines the simulator startup time.
During this period `/health/ready` returns code 503.

Fixes #453


mayabar:
/ready endpoint is marked as deprecated and will be removed in version v0.10.0

PR #485, opened 2026-05-12
Remove support for milliseconds number for configuration latency related parameters 

mayabar:


mayabar:
Co-authored-by: Name @setsunakute

PR #505, opened 2026-05-20
feat: adds `--force-dummy-tokenizer` flag to force the use of the dummy tokenizer regardless of model name.

### LINKED ISSUE #502: Add config flag to force usage of a dummy tokenizer
Currently the simulator has two options for tokenizers: real vllm render which is loaded with the model used by the simulator, and the dummy tokenizer which separates text to tokens using a regular expression.

Add a new configuration parameter (option for name `--force-dummy-tokenizer`) that explicitly overrides the default behavior. When this flag is passed, the system should bypass loading the real tokenizer and instead use a dummy tokenizer, even if a real model name was supplied

sats-23:
Fixes #502 

Currently, the simulator automatically selects the tokenizer based on whether the model name exists on HuggingFace. This means users who want to use real model names for testing but avoid tokenizer loading overhead must use fake model names. This flag provides explicit control over tokenizer selection.

Changes:
- Added `ForceDummyTokenizer` field to `Configuration` struct
- Added `--force-dummy-tokenizer` command-line flag
- Updated tokenizer initialization logic to check flag before model existence
- Updated documentation in `docs/tokenization.md` and `docs/configuration.md`

Usage:
```bash
# Use dummy tokenizer with a real model name
./bin/llm-d-inference-sim --model meta-llama/Llama-3.1-8B-Instruct --force-dummy-tokenizer

### DIFF
--- a/pkg/tests/simulator_test.go
+++ b/pkg/tests/simulator_test.go
@@ -2664,28 +2664,6 @@ var _ = Describe("Simulator", func() {
 				Expect(simulator.Context.Tokenizer).To(BeAssignableToTypeOf(&tokenizer.SimpleTokenizer{}))
 			})
 
-			It("should use dummy tokenizer for text completions when flag is set", func() {
-				ctx := context.TODO()
-				// Use a real model name but force dummy tokenizer
-				args := []string{"cmd", "--model", common.QwenModelName, "--mode", common.ModeEcho, "--force-dummy-tokenizer"}
-				simulator, _, _, err := startServerHandle(ctx, "", args, nil)
-				Expect(err).NotTo(HaveOccurred())
-
-				// Verify that the dummy tokenizer was actually created
-				Expect(simulator.Context.Tokenizer).To(BeAssignableToTypeOf(&tokenizer.SimpleTokenizer{}))
-			})
-
-			It("should use dummy tokenizer for streaming when flag is set", func() {
-				ctx := context.TODO()
-				// Use a real model name but force dummy tokenizer
-				args := []string{"cmd", "--model", common.QwenModelName, "--mode", common.ModeRandom, "--force-dummy-tokenizer"}
-				simulator, _, _, err := startServerHandle(ctx, "", args, nil)
-				Expect(err).NotTo(HaveOccurred())
-
-				// Verify that the dummy tokenizer was actually created
-				Expect(simulator.Context.Tokenizer).To(BeAssignableToTypeOf(&tokenizer.SimpleTokenizer{}))
-			})
-
 			It("should work with YAML config file", func() {
 				ctx := context.TODO()
 				// Create a temporary config file with force-dummy-tokenizer set
--- a/pkg/tests/simulator_test.go
+++ b/pkg/tests/simulator_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/llm-d/llm-d-inference-sim/pkg/dataset"
 	kvcache "github.com/llm-d/llm-d-inference-sim/pkg/kv-cache"
 	openaiserverapi "github.com/llm-d/llm-d-inference-sim/pkg/openai-server-api"
+	"github.com/llm-d/llm-d-inference-sim/pkg/tokenizer"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by mayabar** [comment]
@sats-23 Thanks for your PR!

The PR looks good, please add test for this command line parameter

**Comment by mayabar** [comment]
@sats-23 thanks for the quick fix. Just to clarify - how do the current tests ensure the dummy tokenizer gets created?
In addition, please run `make test` locally.

**Comment by sats-23** [comment]
> @sats-23 thanks for the quick fix. Just to clarify - how do the current tests ensure the dummy tokenizer gets created? In addition, please run `make test` locally.

1. Implicit Creation
-Tests using `common.TestModelName` automatically trigger dummy tokenizer creation because `tokenizer.New()` checks if the model exists and falls back to `NewSimpleTokenizer()` for non-existent models.

2. Explicit Creations (using `--force-dummy-tokenizer`)
-Forces dummy tokenizer creation even with real model names like `common.QwenModelName`. The flag is checked first in `tokenizer.New()`.

& I've also run `make test` locally and verified to be passing!

**Comment by mayabar** [comment]
My concern is about the test assertions. While the tests exercise the right code paths, they don't actually check the outcome. We need to explicitly verify the type of the created tokenizer to ensure the dummy one was actually created, rather than just checking that it didn't return an error. Could you add a check for this using `startServerHandle`? Requests sending is not required in this case, only simulator creation.

**Comment by sats-23** [comment]
> My concern is about the test assertions. While the tests exercise the right code paths, they don't actually check the outcome. We need to explicitly verify the type of the created tokenizer to ensure the dummy one was actually created, rather than just checking that it didn't return an error. Could you add a check for this using `startServerHandle`? Requests sending is not required in this case, only simulator creation.

Hey @mayabar 
I have updated the tests to verify the instance type of the tokenizer, PTAL

**Comment by mayabar** [comment]
> Hey @mayabar I have updated the tests to verify the instance type of the tokenizer, PTAL
@sats-23 Tests failed, please run `make test` locally

**Comment by sats-23** [comment]
Apologies for the delay, the failures have been addressed

File: pkg/tests/simulator_test.go [hunk 1]
```diff
@@ -2650,5 +2651,87 @@ var _ = Describe("Simulator", func() {
...
+			It("should use dummy tokenizer for text completions when flag is set", func() {
+				ctx := context.TODO()
+				// Use a real model name but force dummy tokenizer
+				args := []string{"cmd", "--model", common.QwenModelName, "--mode", common.ModeEcho, "--force-dummy-tokenizer"}
+				simulator, _, _, err := startServerHandle(ctx, "", args, nil)
+				Expect(err).NotTo(HaveOccurred())
+
+				// Verify that the dummy tokenizer was actually created
+				Expect(simulator.Context.Tokenizer).To(BeAssignableToTypeOf(&tokenizer.SimpleTokenizer{}))
+			})
```
mayabar: mode (random/echo) does not affect the Tokenizer, this test if redandunt

pkg/tests/simulator_test.go
```diff
@@ -2650,5 +2651,87 @@ var _ = Describe("Simulator", func() {
...
+			It("should use dummy tokenizer for streaming when flag is set", func() {
+				ctx := context.TODO()
+				// Use a real model name but force dummy tokenizer
+				args := []string{"cmd", "--model", common.QwenModelName, "--mode", common.ModeRandom, "--force-dummy-tokenizer"}
+				simulator, _, _, err := startServerHandle(ctx, "", args, nil)
+				Expect(err).NotTo(HaveOccurred())
+
+				// Verify that the dummy tokenizer was actually created
+				Expect(simulator.Context.Tokenizer).To(BeAssignableToTypeOf(&tokenizer.SimpleTokenizer{}))
+			})
```
mayabar: Tokenizer creation is not related to streaming
The test is same as the first one, please remove

sats-23: Sure, removed it

pkg/tests/simulator_test.go
sats-23: Sure, removed it

PR #510, opened 2026-05-26
Support tokens and arrays of tokens in completions

### LINKED ISSUE #504: Add support for tokens as input in /completions
Support both cases - a single array of token ids, an array of arrays of token ids

irar2:
Fixes #504

### DIFF
--- a/pkg/dataset/ds_tool.go
+++ b/pkg/dataset/ds_tool.go
@@ -215,7 +215,7 @@ func (dt *DatasetTool) conversationToOutputRecords(userTxt, assistantTxt string,
 
 	// create completions request
 	textRequest := openaiserverapi.TextCompletionsRequest{
-		Prompt: userTxt,
+		Prompt: openaiserverapi.PromptInput{Text: userTxt},
 	}
 
 	// add current user message
@@ -225,12 +225,12 @@ func (dt *DatasetTool) conversationToOutputRecords(userTxt, assistantTxt string,
 	})
 
 	// create db record for /completions (without the messages concatenation)
-	inputTokens, _, err := dt.tokenizer.RenderText(textRequest.Prompt)
+	inputTokens, _, err := dt.tokenizer.RenderText(textRequest.Prompt.Text)
 	if err != nil {
-		return nil, errors.Join(err, fmt.Errorf("input tokenization failed (%s)", textRequest.Prompt))
+		return nil, errors.Join(err, fmt.Errorf("input tokenization failed (%s)", textRequest.Prompt.Text))
 	}
 
-	if rec, err := dt.createOutputRecord(inputTokens, assistantTxt, textRequest.Prompt); err != nil {
+	if rec, err := dt.createOutputRecord(inputTokens, assistantTxt, textRequest.Prompt.Text); err != nil {
 		return nil, err
 	} else {
 		result = append(result, *rec)
--- a/pkg/dataset/ds_tool.go
+++ b/pkg/dataset/ds_tool.go
@@ -213,11 +213,6 @@ func (dt *DatasetTool) conversationToOutputRecords(userTxt, assistantTxt string,
 	chatRequest *openaiserverapi.ChatCompletionsRequest) ([]outputRecord, error) {
 	result := []outputRecord{}
 
-	// create completions request
-	textRequest := openaiserverapi.TextCompletionsRequest{
-		Prompt: openaiserverapi.PromptInput{Text: userTxt},
-	}
-
 	// add current user message
 	chatRequest.Messages = append(chatRequest.Messages, openaiserverapi.Message{
 		Role:    openaiserverapi.RoleUser,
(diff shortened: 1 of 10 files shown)


pkg/openai-server-api/request.go
```diff
@@ -495,17 +511,86 @@ func (t *TextCompletionsParsedRequest) UnmarshalJSON(data []byte) error {
...
+		return nil
+	}
+	// prompt is an array of arrays of token ids
+	var tokenLists [][]uint32
+	if err := json.Unmarshal(a.Prompt, &tokenLists); err != nil {
+		return fmt.Errorf("prompt must be a string, an array of strings, an array of token ids, or an array of arrays of token ids: %w", err)
+	}
+	t.Prompt = make([]PromptInput, len(tokenLists))
+	for i, ids := range tokenLists {
+		t.Prompt[i] = PromptInput{Tokens: ids}
```
mayabar, line 542: do we need to check all elements to decide that this is array of strings or tokens
valid request prompt will contain homogenius elements, all of them are tokens and all are strings

pkg/openai-server-api/request.go
```diff
@@ -495,17 +511,86 @@ func (t *TextCompletionsParsedRequest) UnmarshalJSON(data []byte) error {
...
+			for i, p := range t.Prompt {
+				arrs[i] = p.Tokens
+			}
+			a.Prompt = arrs
+		} else {
+			strs := make([]string, len(t.Prompt))
+			for i, p := range t.Prompt {
+				strs[i] = p.Text
+			}
+			a.Prompt = strs
```
mayabar: same

pkg/dataset/ds_tool.go
```diff
@@ -215,7 +215,7 @@ func (dt *DatasetTool) conversationToOutputRecords(userTxt, assistantTxt string,
 
 	// create completions request
 	textRequest := openaiserverapi.TextCompletionsRequest{
-		Prompt: userTxt,
+		Prompt: openaiserverapi.PromptInput{Text: userTxt},
```
mayabar: related to the code before your change - do we really need textRequest, only text is used for rendering, consider to remove the request

pkg/llm-d-inference-sim/text_completion.go
```diff
@@ -40,6 +41,15 @@ func (t *TextCompletionsParsedRequest) validate(_ *toolsValidator) (string, int)
...
 	}
+	for _, p := range t.Prompt {
+		if p.IsTokens() {
+			if len(p.Tokens) == 0 {
+				return "prompt must not contain an empty token-id array", fasthttp.StatusBadRequest
+			}
+		} else if p.Text == "" {
+			return "prompt must not contain an empty string", fasthttp.StatusBadRequest
+		}
+	}
```
mayabar, line 52: consider adding check that all elements in t.Prompt are of the same type

pkg/llm-d-inference-sim/text_completion.go
```diff
@@ -144,6 +158,20 @@ func (t *textCompletionReqCtx) createToolCalls() ([]openaiserverapi.ToolCall, in
...
+		// a trailing comma on all but the last so joining the Strings with ""
+		// yields "id0,id1,...,idN" — what echo mode replays back to the client
+		ids := t.req.Prompt.Tokens
+		strs := make([]string, len(ids))
+		for i, id := range ids {
+			strs[i] = strconv.FormatUint(uint64(id), 10)
+			if i < len(ids)-1 {
+				strs[i] += ","
+			}
+		}
```
mayabar, line 172: maybe create array of strings for numbers and use strings.Join to concat them

pkg/openai-server-api/request.go
irar2: This is not a check, it's unmarshalling. We can do it another way, and just assume that all the elements are of the same type, but we obviously don't want that. And your later comment actually asks to check this.

pkg/llm-d-inference-sim/text_completion.go
irar2: Checked in unmarshal

pkg/llm-d-inference-sim/text_completion.go
irar2: We don't want to concatenate here, we want to have a separate element for each token

pkg/dataset/ds_tool.go
irar2: done

pkg/openai-server-api/request.go
mayabar: I see now that this is unmarshaling, and array of non-homogeneus type will fail to unmarshal. so actually no need to check this not here not in other places, isn't it?

pkg/openai-server-api/request.go
mayabar: maybe here just check the first element if it is a string or tokens and behave accordingly or in a loop check each elements for its type and set according the type

pkg/llm-d-inference-sim/text_completion.go
mayabar: I mean convert [1, 2, 3] to ["1", "2", "3"] and join them to a single string using .Join function instead of strs[i] = strconv.FormatUint(uint64(id), 10) and strs[i] += ","

mayabar: irrelenant

pkg/openai-server-api/request.go
irar2: done

PR #535, opened 2026-06-04
Expose latency fields via /admin/config

irar2:
Part of #520 

All latency-related configuration fields (time-to-first-token, inter-token-latency, prefill-overhead, kv-cache-transfer-latency, time-factor-under-load, latency-calculator, and their std-dev counterparts) can now be updated at runtime via POST /admin/config without restarting the simulator. The latency calculator is atomically rebuilt after each such update; non-latency updates skip the rebuild.

Duration fields accept Go duration strings (e.g. "250ms") in addition to nanosecond integers. The same validation rules as at startup apply; rejected updates leave the calculator unchanged.

Internally, the static adminConfigurableFields map is replaced by struct tags (admin:"configurable", latency:"rebuild", admin:"omit") with a reflection-driven init() — no list to keep in sync with the struct. LoraModulesString is tagged admin:"omit" so it is stripped from GET /admin/config responses automatically.


pkg/common/config.go
```diff
@@ -576,32 +578,114 @@ func (c *Configuration) SSLEnabled() bool {
...
+// rebuilding the latency calculator (admin:"configurable" latency:"rebuild" tags).
+// otherConfigurableFields holds the remaining admin-configurable keys
+// (admin:"configurable" only). omitFields holds keys that are stripped from
+// the external /admin/config response (admin:"omit"). All four maps are populated
+// once at init via reflection so there is no static list to keep in sync with
+// the struct.
+var (
+	durationFields          map[string]bool
+	latencyConfigKeys       map[string]bool
+	otherConfigurableFields map[string]bool
```
mayabar: consider to keep one structure for all configurable field names
it could be a map with key = "rebuild category"/"default (does not cause any rebuld)" 
value is a collection of field names

pkg/common/config_test.go
```diff
@@ -840,3 +888,56 @@ var _ = Describe("Configuration.Copy", func() {
...
+	It("latencyConfigKeys contains the expected latency fields", func() {
+		for _, key := range []string{
+			"time-to-first-token", "time-to-first-token-std-dev",
+			"inter-token-latency", "inter-token-latency-std-dev",
+			"kv-cache-transfer-latency", "kv-cache-transfer-latency-std-dev",
+			"prefill-overhead", "prefill-time-per-token", "prefill-time-std-dev",
+			"kv-cache-transfer-time-per-token", "kv-cache-transfer-time-std-dev",
+			"time-factor-under-load", "latency-calculator",
+		} {
+			Expect(latencyConfigKeys).To(HaveKey(key))
```
mayabar: consider check that latencyConfigKeys does not contain additional keys

pkg/common/config_test.go
```diff
@@ -840,3 +888,56 @@ var _ = Describe("Configuration.Copy", func() {
...
+		}
+	})
+
+	It("otherConfigurableFields contains the expected configurable fields", func() {
+		for _, key := range []string{"failure-injection-rate", "failure-types", "fake-metrics"} {
+			Expect(otherConfigurableFields).To(HaveKey(key))
```
mayabar: same as above

pkg/llm-d-inference-sim/context.go
```diff
@@ -231,15 +260,15 @@ func (s *SimContext) simulateTTFT(respCtx ResponseContext) {
 		DoRemotePrefill:    respCtx.doRemotePrefill(),
 		RunningReqs:        s.metrics.nRunningReqs,
 	}
-	ttft := s.latencyCalculator.GetTimeToFirstToken(&params)
+	ttft := s.LatencyCalc().GetTimeToFirstToken(&params)
```
mayabar: can LatencyCalc be private?

pkg/tests/latencies_test.go
```diff
@@ -67,3 +69,221 @@ var _ = Describe("Check latency calculator", Ordered, func() {
...
+	)
+
+	AfterEach(func() {
+		if ctx != nil {
+			ctx.Done()
+		}
+	})
+
+	Context("config validation and round-trip", func() {
+		BeforeEach(func() {
```
mayabar: consider to change to BeforeAll

PR #536, opened 2026-06-07
pkg: add n parameter for multiple completion choices

umakantv:
Add support for the OpenAI 'n' parameter which controls the number of completion choices generated per request. Each choice is independently generated with different content.

This PR solves https://github.com/llm-d/llm-d-inference-sim/issues/249 and already addresses review feedback on PR #262

Implementation details:
- split() returns n copies of the request for chat/text completions
- For array-prompt text completions, total sub-requests = prompts * n
- aggregateUsage() counts prompt tokens once per unique request ID, correctly handling all combinations: single prompt + n, array prompts, and array prompts + n
- Chat completion response builder updated to produce per-choice messages with independent logprobs

### DIFF
--- a/pkg/communication/response_builder.go
+++ b/pkg/communication/response_builder.go
@@ -83,21 +83,40 @@ type responseBuilder interface {
 	createRenderResponse(tokens [][]uint32, features *openaiserverapi.RenderMMFeatures) any
 }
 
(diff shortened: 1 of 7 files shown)


umakantv:
Closed this PR due to unsigned commits. Will Recreate

PR #548, opened 2026-06-14
feat: add support for image and audio content types in Responses API

### LINKED ISSUE #463: Support additional content types in Responses API (e.g., images)

umakantv:
Fixes https://github.com/llm-d/llm-d-inference-sim/issues/463

Extends the /v1/responses endpoint to accept input_image and input_audio content blocks in addition to the existing input_text, aligning the simulator with the OpenAI Responses API spec for multimodal inputs.


umakantv:
@irar2 @mayabar apart from accepting multimodal input, do we need any other effects to simulate actual processing of the media input?

pkg/llm-d-inference-sim/responses.go
```diff
@@ -87,19 +87,38 @@ func convertInputToMessages(input []openaiserverapi.InputItem) []openaiserverapi
...
+			// Check if content contains only a single text block
+			allText := true
+			for _, content := range inputMsg.Content {
+				if content.Type != openaiserverapi.ResponsesInputText {
+					allText = false
+					break
+				}
+			}
+
+			if allText && len(inputMsg.Content) == 1 {
```
mayabar: In case inputMsg.Content contains more than one item, msg.Content.Raw cannot be used, so there is no need in the loop above.

docs/http-enpdpoints.md
```diff
@@ -174,6 +174,54 @@ Structure of requests/responses
...
                 - offset (token index where the multimodal region begins)
                 - length (number of tokens the region spans)
+- `/v1/responses`
+    - **request**
+        - stream
+        - model
+        - input (array of input items)
+            - type (`message`)
+            - role (`user`, `system`, `developer`)
+            - content (array of content blocks)
```
mayabar: In the request json the 'content' can be not only an array of blocks, but also a string

docs/http-enpdpoints.md
```diff
@@ -174,6 +174,54 @@ Structure of requests/responses
...
+                - bytes
+                - top_logprobs
+        - text
+          - format
+            - type
+        - usage
+          - input_tokens
+          - output_tokens
+          - total_tokens
+        - top_logprobs
```
mayabar: top_logprobs is part of output/content, isn't it?

mayabar (commented):
@umakantv Thanks for your contribution, some comments regarding this PR

pkg/llm-d-inference-sim/responses.go
umakantv: Got it, removed the loop and just used `inputMsg.Content[0].Type == openaiserverapi.ResponsesInputText`

PR #568, opened 2026-06-26
feat: add automatic TLS certificate reloading via fsnotify

pierDipi:
Reuse the CertReloader from llm-d-router to watch the cert directory and hot-reload tls.crt/tls.key on change, so rotated certificates are picked up without a server restart. Falls back to static certs (with a warning) when filenames differ from the Kubernetes convention or when cert and key live in different directories.


shmuelk:
@pierDipi Thank you for your PR.

While I appreciate your desire to re-use code from the llm-d-router repo, I don't think making the llm-d-router a dependency of the llm-d-inference-sim is a good idea.

Furthermore upgrading the version llm-d-kv-cache used by the llm-d-inference-sim often entails API changes and should really be a PR all by itself.

mayabar:
I think it’s important to distinguish between the router and the simulator here. Updating certificates in the router is a production concern, where the component is expected to run continuously without restarts. In contrast, the simulator is a testing tool and is not designed or expected to run indefinitely. Because of that, handling certificate updates in the simulator does not need to follow the same constraints or mechanisms as in the production router. 

@shmuelk what do you think?

shmuelk:
@mayabar wrote:

> I think it’s important to distinguish between the router and the simulator here. Updating certificates in the router is a production concern, where the component is expected to run continuously without restarts. In contrast, the simulator is a testing tool and is not designed or expected to run indefinitely. Because of that, handling certificate updates in the simulator does not need to follow the same constraints or mechanisms as in the production router.

I agree.

pierDipi:
@shmuelk @mayabar removed the router dependency, it's now self-contained in the simulator

pierDipi:
@shmuelk @mayabar please take a look

pkg/common/certs.go
```diff
@@ -0,0 +1,110 @@
...
+
+	go func() {
+		defer func() { _ = w.Close() }()
+
+		var debounceTimer *time.Timer
+
+		for {
+			select {
+			case ev := <-w.Events:
+				traceLogger.Info("Cert changed", "event", ev)
```
mayabar, line 75: Since this action is not supposed to happen in a high rate, I would use logger instead of trace logger

mayabar (approved):

Issue #571, opened 2026-06-29
Simulating very large context lengths fails with HTTP error


jtechapps:
**What happened**:

I am trying to simulate DeepSeek-V4-Flash with 1M token context window.  I am able to run simulator pods fine using the following configuration:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: llm-d-sim
  name: llm-d-sim
  namespace: llm-d-sim
spec:
  replicas: 10
  selector:
    matchLabels:
      app: llm-d-sim
  template:
    metadata:
      labels:
        app: llm-d-sim
    spec:
      topologySpreadConstraints:
      - maxSkew: 1
        topologyKey: topology.kubernetes.io/zone
        whenUnsatisfiable: DoNotSchedule
        labelSelector:
          matchLabels:
            app: llm-d-sim
      containers:
      - args:
        - --model
        - deepseek-ai/DeepSeek-V4-Flash
        - --port
        - '8000'
        - --max-model-len
        - '1000000'
        - --max-num-seqs
        - '256'
        - --max-waiting-queue-length
        - '2048'
        - --inter-token-latency
        - 1ms
        - --prefill-time-per-token
        - 4us
        - --prefill-overhead
        - 5ms
        - --enable-kvcache
        - --kv-cache-size
        - '10000000'
        - --block-size
        - '16'
        - --force-dummy-tokenizer
        env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: POD_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
        image: ghcr.io/llm-d/llm-d-inference-sim:v0.10.0
        imagePullPolicy: IfNotPresent
        livenessProbe:
          httpGet:
            path: /health/ready
            port: 8000
          initialDelaySeconds: 15
          periodSeconds: 10
        name: vllm-sim
        ports:
        - containerPort: 8000
          name: http
          protocol: TCP
        - containerPort: 5557
          name: zmq-kv-events
          protocol: TCP
        readinessProbe:
          httpGet:
            path: /health/ready
            port: 8000
          initialDelaySeconds: 10
          periodSeconds: 5
        resources:
          requests:
            cpu: '1'
            memory: 8Gi
```

I am simulating agentic inference requests using inference-perf jobs with the following data generation:
```
[...]

PR #572, opened 2026-06-30
feat: add max-request-body-size-mb configuration flag (#571)

irar2:
Adds a `--max-request-body-size-mb` flag to cap the HTTP server's maximum request body size. Default is 4 MB (matching the fasthttp built-in default), valid range is 1–512 MB.


shmuelk (approved):

jtechapps:
Fixes: https://github.com/llm-d/llm-d-inference-sim/issues/571

PR #588, opened 2026-07-22
feat: add ZMQ replay endpoint for KV cache events

### LINKED ISSUE #477: Support kvevent replay
Support kvevent replay via ZMQ ROUTER socket.

**Why is this needed**:
vLLM provides kvevent replay mechanism to eliminate the sequence gap when kvevent message missed.

irar2:
## What does this PR do?

Adds an optional ROUTER-socket replay endpoint (`--kv-events-replay-endpoint`) that lets clients request missed KV cache event batches by sequence number, matching vLLM's replay wire format (topic/seq/payload framing, end-of-replay sentinel). Published batches are kept in a bounded, configurable ring buffer (`--kv-events-replay-queue-size`, default 1024) and offset by data-parallel rank the same way as the existing ZMQ publish endpoint, with validation to prevent the two endpoints' ports from colliding across ranks.

## Related Issues

Fixes #477

### DIFF
--- a/pkg/common/config.go
+++ b/pkg/common/config.go
@@ -222,6 +222,13 @@ type Configuration struct {
 	// ZMQEndpoint is the ZMQ address to publish events, the default value is tcp://localhost:5557
 	ZMQEndpoint string `yaml:"zmq-endpoint" json:"zmq-endpoint"`
 
+	// KVEventsReplayEndpoint is the ZMQ PULL address to bind for receiving KV events replay requests.
+	// Empty (default) disables the replay listener. Example: "tcp://*:5558"
+	KVEventsReplayEndpoint string `yaml:"kv-events-replay-endpoint" json:"kv-events-replay-endpoint"`
+
+	// KVEventsReplayQueueSize is the max number of event batches held in the replay queue; oldest dropped when full. Defaults to 1024.
+	KVEventsReplayQueueSize int `yaml:"kv-events-replay-queue-size" json:"kv-events-replay-queue-size"`
+
 	// EventBatchSize is the maximum number of kv-cache events to be sent together, defaults to 16
 	EventBatchSize int `yaml:"event-batch-size" json:"event-batch-size"`
 
@@ -364,6 +371,7 @@ func newConfig() *Configuration {
 		KVCacheSize:                1024,
 		TokenBlockSize:             16,
 		ZMQEndpoint:                "tcp://127.0.0.1:5557",
+		KVEventsReplayQueueSize:    1024,
 		EventBatchSize:             16,
 		DPSize:                     1,
 		Rank:                       -1,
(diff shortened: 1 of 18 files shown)


pkg/common/config.go
```diff
@@ -626,6 +642,38 @@ func (c *Configuration) validate() error {
...
+		return nil
+	}
+
+	// When data-parallel-rank is set, data-parallel-size is ignored for endpoint
+	// offsetting and both endpoints are offset by the
+	// same fixed rank, so only an exact port match can collide.
+	maxRank := c.DPSize - 1
+	if c.Rank >= 0 {
+		maxRank = 0
+	}
```
mayabar: when dp size and rank are defined, we need to check that the distance between zmqPort and replayPort is larger than the dp size, without relation to current instance dp rank

pkg/kv-cache/kv_events_replayer_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	zmq4 "github.com/go-zeromq/zmq4"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// sendReplayRequestAndRecv connects a REQ socket to the replayer ROUTER endpoint,
+// sends an 8-byte big-endian start sequence number, and returns all reply frames
+// received until the end-of-replay sentinel (seq == 0xFFFFFFFFFFFFFFFF).
+func sendReplayRequestAndRecv(ctx context.Context, endpoint string, startSeq uint64) []zmq4.Msg {
```
mayabar: consider to use same utility as in simulator test

pkg/kv-cache/kv_events_replayer_test.go
```diff
@@ -0,0 +1,205 @@
...
+		if seq == replaySentinel {
+			break
+		}
+	}
+	return replies
+}
+
+var _ = Describe("kvEventsReplayer", func() {
+	const (
+		replayEndpoint = "tcp://127.0.0.1:15558"
```
mayabar: maybe use any free port?

pkg/kv-cache/kv_events_replayer_test.go
```diff
@@ -0,0 +1,205 @@
...
+			r.store(3, []byte("batch-3"))
+
+			Expect(r.queue.len()).To(Equal(3))
+			entries := r.queue.since(2)
+			Expect(entries).To(HaveLen(2))
+			Expect(entries[0].seq).To(Equal(uint64(2)))
+			Expect(entries[0].payload).To(Equal([]byte("batch-2")))
+		})
+
+		It("slides out old batches when queue is full", func() {
```
mayabar: consider to remove - redundant

mayabar (approved):

PR #597, opened 2026-07-29
fix a few infinite loop issues

yamt:
## What does this PR do?

fix a few infinite loop issues when hitting max-model-len.

## Why is this change needed?

the infinite loop makes a worker stop working ~forever.

## How was this tested?

- [ ] Unit tests added/updated

## Checklist

### DIFF
--- a/docs/response_generation.md
+++ b/docs/response_generation.md
@@ -55,7 +55,7 @@ If `max_tokens` or `max_completion_tokens` is specified, the response length is 
 |6|20|181-200|
 
 **Default Length:**
-If no maximum length is specified, the length defaults to `<model_context_limit> - <input_length>`. In this specific case, the length is sampled from a Gaussian distribution (Mean=40, SD=20).
+If no maximum length is specified, the length defaults to `<model_context_limit> - <input_length>`. In this specific case, unless the calculated maximum length is smaller than 20, the length is sampled from a Gaussian distribution (Mean=40, SD=20). Otherwise, the length is chosen uniformly at random.
 
 ### Content Generation Source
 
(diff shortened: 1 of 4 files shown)


irar2:
@yamt Please sign off the commits

mayabar:
@yamt thanks for catching this. After reviewing how response length is calculated using a Gaussian with mean 40 and standard deviation 20, I think the current behavior could be improved: when `maxLen` is below 20 (i.e. `mean - stddev`), we should skip the Gaussian and pick a value uniformly at random from `[1, maxLen]` instead.
Can you update this PR to this behavior please?

yamt:
> @yamt thanks for catching this. After reviewing how response length is calculated using a Gaussian with mean 40 and standard deviation 20, I think the current behavior could be improved: when `maxLen` is below 20 (i.e. `mean - stddev`), we should skip the Gaussian and pick a value uniformly at random from `[1, maxLen]` instead. Can you update this PR to this behavior please?

well, what's the original motivation to use Gaussian distribution?
is it ok to switch to a different distribution silently?

mayabar:
The Gaussian distribution was chosen based on research about expected response length, which showed that responses are typically around 40 tokens. I think picking a random value for `maxLen` less than 20 is acceptable.
If you have concerns, you can add an explanation of the new behavior to the documentation (response_generation.md).

yamt:
> @yamt Please sign off the commits

done

yamt:
> @yamt thanks for catching this. After reviewing how response length is calculated using a Gaussian with mean 40 and standard deviation 20, I think the current behavior could be improved: when `maxLen` is below 20 (i.e. `mean - stddev`), we should skip the Gaussian and pick a value uniformly at random from `[1, maxLen]` instead. Can you update this PR to this behavior please?

done

mayabar (approved):

PR #598, opened 2026-07-29
fix: report real HTTP status for streaming requests that fail before producing content

### LINKED ISSUE #596: openai api streaming incompatibility with vllm when hitting max_model_len
### What happened?

when hitting max_model_len, the sim's openai api (v1/chat/completions) with stream=true 
returns error within a stream:

```
{'error': {'message': "This model's maximum context length is 1024 tokens. Howev
er, you requested 1029 tokens (1028 in the messages, 1 in the completion). Pleas
e reduce the length of the messages or completion", 'type': 'BadRequestError', '
param': None, 'code': 400}}
```

it's incompatible with the real vllm, which returns http status 400 
in that case.

### How to reproduce it (as minimally and precisely as possible)

run llm-d-inference-sim with a possibly small --max-model-len, and send a large prompt which doesn't fit it via open ai api with stream=true.

### Anything else we need to know?

_No response_

### Environment

_No response_

irar2:
## Summary
- Streaming requests always returned HTTP 200 even when the request failed
  before any content was produced (e.g., queue full, or a validation error
  only caught once processing starts). This is because `sendStream` set
  `SetStatusCode(200)` and handed the response to `SetBodyStream` before
  knowing whether the first message on the channel would be an error.
- `handleHTTP` now peeks the first response before committing to a stream.
  If it's an error, we drain the channel and call `sendError` with the real
  status code - no SSE stream is started. Otherwise we commit to 200 +
  `text/event-stream` and feed the already-consumed first message into
  `sendStream` so it isn't dropped.
- Also fixed `emitResponseChunks`'s defensive "unexpected response part in
  streaming" branch to use `sendStreamErrorAndDone` instead of `sendError`,
  since it runs inside the same streaming goroutine where the status code is
  already committed - `sendError` there was a no-op/race, not an actual
  fix-up of the response.

## Notes
- This only helps when the error is the very first message observed
  (single-choice requests). The multi-choice fail-fast case, where one
  choice has already streamed tokens before another choice errors, is
  unchanged - it's still correctly surfaced as an in-band SSE error frame
  followed by `[DONE]`, since the HTTP status is unavoidably already locked
  to 200 by that point.

Fixes #596


mayabar (approved):

yamt (approved):
lgtm. thank you.

PR #604, opened 2026-08-03
fix: bump golangci-lint to v2.10.1 to fix analysis panic

irar2:
`make lint` panics on v2.10.0 and is fixed upstream in v2.10.1.

Change: bump `GOLANGCI_LINT_VERSION` from v2.10.0 to v2.10.1.

### DIFF
--- a/Makefile.tools.mk
+++ b/Makefile.tools.mk
@@ -30,7 +30,7 @@ ln -sf $$(realpath $(1)-$(3)) $(1)
 endef
 
 GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
-GOLANGCI_LINT_VERSION = v2.10.0
+GOLANGCI_LINT_VERSION = v2.10.1
 GINKGO = $(LOCALBIN)/ginkgo
 GINKGO_VERSION = v2.25.3
 


nilig (approved):
Verified the bump does what the title says — `v2.10.1` carries exactly one commit, the `buildssa` panic fix (golangci/golangci-lint#6376), so this is the minimal step that resolves it. LGTM.

Two small things:

We have a second pin at `.github/workflows/ci-pr-checks.yaml:57` (`golangci/golangci-lint-action@v9`, `version: "v2.10.0"`) that the action reads instead of the Makefile, so after this merges local `make lint` and CI would be on different versions. Not urgent — CI can't hit the panic anyway, see below — but I think we want both moving together rather than drifting.

nit: worth a line in the description on why `lint-and-test` is green at v2.10.0, otherwise this reads as a no-op fix. Per the upstream root cause (golangci/golangci-lint#6375) the panic needs a *partially* warm fact cache: an action misses `loadCachedFacts()` and marks `buildssa`, but `markDepsForAnalyzingSource` isn't recursive so `ctrlflow` stays unmarked and `buildssa` asserts on its nil `Result`. On a uniformly cold cache every action marks its own deps, so nothing panics — and CI sets `skip-cache: true` on a fresh runner, so it's structurally in the masking case.

PR #605, opened 2026-08-03
feat(responses): add function_call tool turns

### LINKED ISSUE #602: Feature request: Responses API tool calls for Agentic Loop
### What would you like to be added?

Add non-streaming Responses tool-call support:

- Accept `tools` and `tool_choice` on `/v1/responses` (flat function schema)
- When tools are present and choice is not `none`, return **exactly one** `output` item of type `function_call` (`status: completed`, with `id`, `call_id`, `name`, `arguments`)
- Accept `function_call` and `function_call_output` in `input`
- If `input` already includes `function_call_output`, return a text `message` (no further tool call)

Out of scope: streaming function-call events, parallel tool calls, server-side tool execution (that stays with the external driver / MCP).

### Why is this needed?

Agentic Loop drivers sit in front of the model: they send `/v1/responses`, run tools when the model asks, then call the model again until it returns a final answer.

That only works if the simulator can:
1. Emit a `function_call` when tools are offered
2. Accept `function_call` / `function_call_output` on the next request
3. Return a normal `message` after tool results so the driver can stop

Drivers that use Responses + tools cannot exercise the loop against the sim.

aslakknutsen:
## What does this PR do?

Emit one function_call on the first tool turn, accept function_call_output on re-entry, then return a message so the driver can exit the loop.

## Why is this change needed?

Support Agentic Loop round-trips driven by an external client.

## How was this tested?

## Checklist

- [ ] Code follows project [contributing guidelines](https://github.com/llm-d/.github/blob/main/CONTRIBUTING.md)

## Related Issues

Fixes #602

### DIFF
--- a/docs/http-endpoints.md
+++ b/docs/http-endpoints.md
@@ -269,7 +269,7 @@ Structure of requests/responses
           - output_tokens
           - total_tokens
 
-    Tool turns (non-streaming): when `tools` are present and `tool_choice` is not `none`, the simulator emits exactly one `function_call` output item. If `input` already contains a `function_call_output` (agentic loop re-entry), it returns a normal assistant `message` instead. Parallel / multi tool calls are not supported.
+    Tool turns: when `tools` are present and `tool_choice` is not `none`, the simulator emits exactly one `function_call` output item (non-streaming) or the equivalent SSE event sequence (streaming; see below). If `input` already contains a `function_call_output` (agentic loop re-entry), it returns a normal assistant `message` instead. Parallel / multi tool calls are not supported.
 
 - `/inference/v1/generate`
     - **request**
@@ -416,6 +416,8 @@ curl -N -X POST http://localhost:8000/v1/responses \
 ```
 
 The streaming response uses Server-Sent Events (SSE) and emits the following event types in order: `response.created`, `response.in_progress`, `response.output_item.added`, `response.content_part.added`, one or more `response.output_text.delta`, `response.output_text.done`, `response.content_part.done`, `response.output_item.done`, `response.completed`.
+
+When the turn is a tool call (`tools` present and `tool_choice` is not `none`, and `input` has no `function_call_output`), the stream instead emits: `response.created`, `response.in_progress`, `response.output_item.added` (item type `function_call`), one or more `response.function_call_arguments.delta`, `response.function_call_arguments.done`, `response.output_item.done`, `response.completed` (with a single `function_call` in `output`).
 
 ## `finish_reason` values
 
--- a/docs/http-endpoints.md
+++ b/docs/http-endpoints.md
@@ -269,7 +269,9 @@ Structure of requests/responses
           - output_tokens
           - total_tokens
 
-    Tool turns (non-streaming): when `tools` are present and `tool_choice` is not `none`, the simulator emits exactly one `function_call` output item. If `input` already contains a `function_call_output` (agentic loop re-entry), it returns a normal assistant `message` instead. Parallel / multi tool calls are not supported.- `/inference/v1/generate`
+    Tool turns (non-streaming): when `tools` are present and `tool_choice` is not `none`, the simulator emits exactly one `function_call` output item. If `input` already contains a `function_call_output` (agentic loop re-entry), it returns a normal assistant `message` instead. Parallel / multi tool calls are not supported.
+
+- `/inference/v1/generate`
     - **request**
         - stream
         - model
--- a/docs/http-endpoints.md
+++ b/docs/http-endpoints.md
@@ -234,7 +234,7 @@ Structure of requests/responses
         - instructions
         - max_output_tokens
         - tools (array of function tools; flat Responses shape: `type`, `name`, `description`, `parameters`)
-        - tool_choice (`none`, `auto`, `required`, or `{"type":"function","name":"..."}`)
+        - tool_choice (`none`, `auto`, `required`, `{"type":"function","name":"..."}`, or wire shapes `allowed_tools` / `custom` — the latter two are accepted but not enforced)
         - text
           - format
             - type (`text`, `json_object`, `json_schema`)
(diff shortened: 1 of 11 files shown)


pkg/api/request.go
```diff
@@ -915,28 +991,105 @@ func (req *ResponsesRequest) UnmarshalJSON(data []byte) error {
...
+	tools := make([]Tool, 0, len(raw))
+	for _, r := range raw {
+		var nested Tool
+		if err := json.Unmarshal(r, &nested); err == nil && nested.Function.Name != "" {
+			if nested.Type == "" {
+				nested.Type = "function"
+			}
+			tools = append(tools, nested)
+			continue
+		}
```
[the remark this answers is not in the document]
aslakknutsen: fixed https://github.com/llm-d/llm-d-inference-sim/pull/605/changes/79d8e1886f19c88fd9849cc53a723c6a1656f0fa

docs/http-endpoints.md
```diff
@@ -244,26 +248,28 @@ Structure of requests/responses
...
         - text
           - format
             - type
         - usage
           - input_tokens
           - output_tokens
           - total_tokens
-- `/inference/v1/generate`
+
+    Tool turns (non-streaming): when `tools` are present and `tool_choice` is not `none`, the simulator emits exactly one `function_call` output item. If `input` already contains a `function_call_output` (agentic loop re-entry), it returns a normal assistant `message` instead. Parallel / multi tool calls are not supported.- `/inference/v1/generate`
```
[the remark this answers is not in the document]
aslakknutsen: fixed https://github.com/llm-d/llm-d-inference-sim/pull/605/changes/79d8e1886f19c88fd9849cc53a723c6a1656f0fa

irar2:
@aslakknutsen What happens with requests with `stream: true`?

aslakknutsen:
@irar2 With `stream: true` + tools (and no `function_call_output` in `input`), the simulator now emits the OpenAI Responses tool SSE sequence:

`response.created` → `response.in_progress` → `response.output_item.added` (`function_call`) → one or more `response.function_call_arguments.delta` → `response.function_call_arguments.done` → `response.output_item.done` → `response.completed` (single `function_call` in `output`).

Same single-call policy as non-streaming. On agentic-loop re-entry (`function_call_output` present), streaming stays on the normal message/text event path.

^ oops.. my agent went rouge. 🙈

Added SEE support as well.

pkg/llm-d-inference-sim/responses.go
irar2: This code now skips the new *api.InputFunctionCall / *api.InputFunctionCallOutput items this PR adds. This function feeds encode() → TokenizedPrompt() → getNumberOfPromptTokens(), which is what populates usage.input_tokens (and would feed any prefix/cache-hit-rate logic that depends on the full prompt). 
Please also add a test that checks usage.input_tokens.

irar2: Same problem as above: only *api.InputMessage is handled, so mode: echo combined with a tool round-trip (last item = function_call_output) echoes an empty string instead of the tool output.

docs/http-endpoints.md
```diff
@@ -244,25 +248,29 @@ Structure of requests/responses
...
         - text
           - format
             - type
         - usage
           - input_tokens
           - output_tokens
           - total_tokens
+
+    Tool turns: when `tools` are present and `tool_choice` is not `none`, the simulator emits exactly one `function_call` output item (non-streaming) or the equivalent SSE event sequence (streaming; see below). If `input` already contains a `function_call_output` (agentic loop re-entry), it returns a normal assistant `message` instead. Parallel / multi tool calls are not supported.
+
```
irar2, line 273: Please consider adding an explanation that every first-turn request with tools and tool_choice omitted or "auto" is forced to call a tool 100% of the time, which differs from the real Responses API.

Also, may be worth mentioning that tool-calling is permanently switched off for the rest of the conversation after the first tool call, even for a later, unrelated user question that would legitimately need a tool again.

pkg/llm-d-inference-sim/responses.go
```diff
@@ -127,7 +147,31 @@ func (r *responsesReqCtx) encode() ([]uint32, []string, *api.RenderMMFeatures, e
...
+	if api.HasFunctionCallOutput(req.Input) {
+		return nil, 0, "", nil
+	}
+
+	toolChoice := req.GetToolChoice()
+	// Deterministic first tool turn: treat omitted/auto as required unless a function is forced.
+	if toolChoice.GetFunction() == nil &&
+		(param.IsOmitted(toolChoice.OfAuto) || toolChoice.OfAuto.Or("") == toolChoiceAuto) {
+		toolChoice = api.NewToolChoiceRequired()
+	}
```
irar2, line 190: This condition fires whenever GetFunction() is nil and OfAuto is unset, which is also true for tool_choice: {"type": "allowed_tools", ...} and {"type": "custom", ...} (neither sets OfFunctionToolChoice or OfAuto). Both get silently rewritten to unrestricted required, so the model can pick any tool in the full list, ignoring the allowed subset / custom-tool restriction.

createToolCalls in tools_utils.go already documents that it doesn't implement allowed_tools/custom forcing (lines 105-110) — so the gap itself predates this PR. But for chat completions that gap only produces a possibly-wrong result: createToolCalls is called with the choice unmodified, so allowed_tools/custom fall into the min := 0 branch, which can still legitimately emit zero tool calls. This PR's auto→required override removes that: for Responses, the exact same tool_choice shapes now guarantee an out-of-scope tool call on every request.

We could possibly exclude allowed_tools/custom from this override (leave them as auto-like, min = 0) so Responses doesn't behave strictly worse than chat completions here. And either way, it is worth mentioning in the docs that allowed_tools/custom aren't enforced, since the current docs (docs/http-endpoints.md:237) only list none/auto/required/forced-function and implies nothing else is accepted.

irar2:
Also please consider adding tests for the following scenarios:

- Forced tool_choice naming a function that doesn't exist in tools

- Mode echo combined with a tool round-trip, where the last input item is a function_call_output

- A second, unrelated user turn after a completed tool round-trip in the same conversation - to confirm that tools staying disabled for the rest of the conversation is the intended behavior

pkg/llm-d-inference-sim/responses.go
aslakknutsen: fixed https://github.com/llm-d/llm-d-inference-sim/pull/605/commits/6a5becfada644d6c1b0d7ed74393ac0d0031dc7b

aslakknutsen: fixed https://github.com/llm-d/llm-d-inference-sim/pull/605/commits/6a5becfada644d6c1b0d7ed74393ac0d0031dc7b

docs/http-endpoints.md
aslakknutsen: fixed https://github.com/llm-d/llm-d-inference-sim/pull/605/commits/6a5becfada644d6c1b0d7ed74393ac0d0031dc7b

pkg/llm-d-inference-sim/responses.go
aslakknutsen: fixed https://github.com/llm-d/llm-d-inference-sim/pull/605/commits/6a5becfada644d6c1b0d7ed74393ac0d0031dc7b

PR #606, opened 2026-08-04
fix: bump CI's golangci-lint pin to v2.10.1

irar2:
Following up on Nili’s comments in #604. The previous PR fixed a `golangci-lint` panic by bumping the version in `Makefile.tools.mk`, but left CI on the buggy `v2.10.0`. CI stayed green regardless because its cold, skip-cache: true runs never hit the specific cache-warmth condition that triggers the panic, while local make lint, which reuses a persistent cache, did.

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -54,7 +54,7 @@ jobs:
       - name: Run lint checks
         uses: golangci/golangci-lint-action@v9
         with:
-          version: "v2.10.0"
+          version: "v2.10.1"
           args: "--config=./.golangci.yml"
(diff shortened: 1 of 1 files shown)


nilig (approved):
Confirmed this is the last one — `Makefile.tools.mk:33` is already on `v2.10.1` and this is the only remaining `v2.10.0` in the repo, so the skew closes here. LGTM.

nit, non-blocking: `lint-and-test` skipped on this PR because the `check-changes` filter doesn't include `.github/workflows/**`, so the new pin won't actually run until the next source PR. Not worth widening this PR for; just flagging that workflow-only changes go unvalidated today.

PR #608, opened 2026-08-04
feat: accept Claude/Anthropic-shaped tool schemas (skip-validation + type normalization)

### LINKED ISSUE #607: Allow bypassing the built-in tool-schema validator (real vLLM does not meta-validate incoming tool schemas)
## Summary

`llm-d-inference-sim` validates every incoming tool's `function.parameters` against a strict, hard-coded JSON Schema in [`pkg/llm-d-inference-sim/tools_utils.go`](https://github.com/llm-d/llm-d-inference-sim/blob/v0.9.0/pkg/llm-d-inference-sim/tools_utils.go#L316) that only whitelists a small subset of fields (`type`, `description`, `enum`, `properties`, `items`, `required`, `additionalProperties`, `minItems`, `maxItems`) with `additionalProperties: false`. Real vLLM does not meta-validate tool schemas this way — it forwards them to the model verbatim. As a result, any client that sends tool schemas with fields outside this whitelist gets a 400 from the sim while the same request passes through real vLLM.

Please add a `--skip-tool-validation` flag (also exposed via YAML config as `skip-tool-validation: true`) that lets deployments opt out of the built-in validator and match real-vLLM behavior. Default `false` preserves current behavior.

## Why this matters

The sim is positioned as a drop-in replacement for testing infrastructure without a GPU. The moment a client sends tools — which any OpenAI-compatible client with tool-calling can — the drop-in illusion breaks:

- **Anthropic API translators** (Claude Code, LiteLLM Anthropic mode, LangChain's Anthropic adapter) convert `input_schema` verbatim and routinely include `default`, `format`, `title`, `pattern`, `minLength`, `maximum`, `oneOf`, `$ref`, `examples`, and `nullable` in property definitions — none of which are in the sim's whitelist.
- **OpenAI's own function-calling docs and cookbook** show tool schemas that include `default` and `format` in properties — legal JSON Schema draft-07 that real vLLM accepts.
- **Structured-output client SDKs** (Zod → JSON Schema, Pydantic → JSON Schema, TypeBox) emit richer JSON Schemas by default. All of them are rejected today.

## Reproducer

Sim v0.9.0:
```bash
docker run --rm -p 8000:8000 ghcr.io/llm-d/llm-d-inference-sim:v0.9.0 \
  --model gemma-4 --port 8000 --startup-duration 1s
```

Request with a tool schema containing `default` and `minLength` (both valid JSON Schema draft-07):
```bash
curl -sS -X POST http://localhost:8000/v1/chat/completions \
  -d '{
    "model":"gemma-4",
    "messages":[{"role":"user","content":"hi"}],
    "tools":[{
      "type":"function",
      "function":{
        "name":"greet",
        "description":"say hi",
        "parameters":{
          "type":"object",
          "properties":{
            "name":{"type":"string","description":"who","default":"world","minLength":1}
          },
          "required":["name"]
        }
      }
    }]
  }'
```

**Actual:**
```json
{"error":{"message":"Tool validation failed: jsonschema: '/parameters/properties/name' does not validate with file:///schema.json#/properties/parameters/$ref/properties/properties/additionalProperties: additional properties 'default', 'minLength' not allowed", ...}}
```

**Expected (matches real vLLM):** a normal chat completion; the sim can either ignore the extra fields, or accept them into its random arg generator (existing generator already handles `type` + `properties`; extras just aren't consulted).

## Proposed change

Add a boolean config knob that short-circuits the validator loop. Minimal diff (based on `v0.9.0`, ~15 lines across 6 files):

**`pkg/common/config.go`** — new field alongside `EnableSleepMode`:
```go
// SkipToolValidation, when true, bypasses the built-in JSON Schema
// validation of incoming tool definitions and forwards them to the
// tool-call generator as-is (matches real vLLM behavior). Default false.
SkipToolValidation bool `yaml:"skip-tool-validation" json:"skip-tool-validation"`
```

**`pkg/common/parser.go`** — register flag near the other Bool flags:
```go
f.BoolVar(&config.SkipToolValidation, "skip-tool-validation", config.SkipToolValidation,
    "Skip JSON Schema validation of incoming tool definitions; forward to the tool-call generator as-is (matches real vLLM behavior)")
```

**`pkg/llm-d-inference-sim/request.go`** — extend the interface:
```go
validate(toolsValidator *toolsValidator, skipToolValidation bool) (string, int)
```

**`pkg/llm-d-inference-sim/chat_completion.go`** — gate the loop:
```go
func (c *ChatCompletionsRequest) validate(toolsValidator *toolsValidator, skipToolValidation bool) (string, int) {
    if !skipToolValidation {
        for _, tool := range c.Tools {
            // existing loop unchanged
        }
    }
    return validateRequest(c)
}
```

Four peer files (`generate.go`, `generation.go`, `responses.go`, `text_completion.go`) get the same signature update — for those non-tool paths the flag is unused, they just pass it through.

**`pkg/llm-d-inference-sim/simulator.go:325`** — pass the flag:
```go
errMsg, errCode := req.validate(s.toolsValidator, s.Context.Config.SkipToolValidation)
```

## Verification

With `--skip-tool-validation` set, the reproducer above returns:
```json
{"id":"chatcmpl-…","model":"gemma-4","object":"chat.completion",…,
 "choices":[{"index":0,"finish_reason":"length",
  "message":{"role":"assistant","content":"…"}}]}
```

Backward compat: without the flag, current behavior is byte-identical (same validator, same rejects).

## Alternatives considered

- **Relax the built-in schema to full JSON Schema draft-07/2020-12**: correct long-term but a larger change with more testing surface — the strict schema serves the synthetic tool-call generator which assumes only the whitelisted fields.
- **`--tool-schema-file <path>`**: allow the schema itself to be overridden from disk (nice for ConfigMap-mounted deployments). Could be a follow-up if there's appetite.

I'd be happy to send the PR if this direction sounds acceptable.

## Environment

- llm-d-inference-sim v0.9.0
- Deployed via `metadata.yaml` → `serve.image` in a fleet control plane (launchpad-ai gateway); requests originate from a Claude Code client hitting the gateway's `/v1/messages` (Anthropic) endpoint.

rishi-anand:
Closes #607.

## What this PR does

Five focused fixes to make the sim serviceable behind a real OpenAI/Anthropic client (Claude Code, LiteLLM, LangChain, custom agents) without giving up on realistic behavior. Each commit is self-contained; feel free to squash if that fits your history preferences.

### 1. `--skip-tool-validation` &nbsp;·&nbsp; commit `3d4d862`

The sim runs every incoming tool through a strict inline JSON Schema at [`pkg/llm-d-inference-sim/tools_utils.go:316-408`](https://github.com/llm-d/llm-d-inference-sim/blob/v0.9.0/pkg/llm-d-inference-sim/tools_utils.go#L316-L408) with `additionalProperties: false` — only `type`/`description`/`enum`/`properties`/`items`/`required`/`additionalProperties`/`minItems`/`maxItems` are allowed. Any tool schema containing valid JSON Schema draft-07 fields (`default`, `format`, `title`, `pattern`, `minLength`, `oneOf`, `$ref`, `examples`, …) is rejected 400.

Real vLLM does not meta-validate incoming tool schemas — it forwards them to the model. This flag lets deployments match that behavior; the loop is skipped when the flag is set. Default `false` = zero behavior change.

### 2. Argument-generator normalizes JSON Schema `type` &nbsp;·&nbsp; commit `08c7c43`

Even with `--skip-tool-validation`, the synthetic arg generator's switch on `propertyMap["type"]` only handled a single concrete string type and errored on:
- **`nil`** — property with no `type` field
- **`["null", "object"]`** — the draft-04+ nullable pattern
- **`["null", "array"]`** — same for arrays

New `resolvePrimitiveType()` unwraps arrays to the first non-`"null"` entry, defaults missing type to `"string"`, and treats plain `"null"` as JSON null. Real vLLM tolerates these because it forwards to the model; the sim must reduce them to a concrete type before synthesizing arguments.

### 3. Nil-safe `items` / `properties` in `createArgument` &nbsp;·&nbsp; commit `7fe81a4` (part 1)

The recursive generator did unchecked type assertions:
```go
itemsMap := propertyMap["items"].(map[string]any)              // PANIC if items nil
objectProperties := propertyMap["properties"].(map[string]any) // PANIC if properties nil
```

Both shapes turn up in real tool schemas (e.g. Claude Code's `WebFetch` declares `{"type": "object"}` with no `properties`). Comma-ok assertions now return an empty `{}` / `[]` — matching how real vLLM tolerates under-specified schemas.

### 4. `--tool-choice-override=[none|auto|required]` &nbsp;·&nbsp; commit `7fe81a4` (part 2)

The sim has no model, so its "should I call a tool?" decision in `auto` mode is effectively a dice roll driven by `RandomInt(0, len(tools))`. With a client attaching a full tool suite (~30+ tools), the dice roll almost always picks a tool call — the sim fabricates tool invocations for plain conversational queries.

Real vLLM doesn't have this issue because a real model uses intent understanding. This adds a deployment-level override that ignores the request's `tool_choice`:
- `""` (default) — honor the client's `tool_choice`; behavior unchanged
- `"none"` — never generate tool calls; always return text
- `"auto"` — behave as if the client sent `auto` (the sim decides)
- `"required"` — always generate at least one tool call

### 5. `--max-completion-tokens-cap` &nbsp;·&nbsp; commit `8d245c9`

`DefaultDataset.calculateResponseMaxLen` uses the request's `max_tokens` verbatim, or `max-model-len - prompt_length` when `max_tokens` is unset. With a client that sends huge `max_tokens` (Anthropic's Messages API always sets it), the sim produces very long responses (many hundreds of concatenated random sentences) that read as "infinite."

This adds a hard ceiling: `MaxCompletionTokensCap > 0` clamps the effective max regardless of the request. `CustomDataset` inherits the cap via embedding.

## Bonus: `--log-http` decompresses gzip bodies &nbsp;·&nbsp; commit `1562b96`

`fasthttp` auto-gzips response bodies when the client sends `Accept-Encoding: gzip` (every Prometheus scraper does). The `--log-http` path was writing the raw compressed bytes to the log — operators saw garbled binary in `kubectl logs` instead of the JSON they needed for debugging tool payloads. New `decompressForLog()` transparently gunzips and truncates for the log, symmetric on request and response; raw wire bytes are unchanged.

## Verification

**Unit-level, local build:**
- `go build ./...` clean.
- Manual smoke test with `--tool-choice-override=none` + `tool_choice: "required"` from client + tools declared → response has `finish_reason: "stop"`, `tool_calls: null`, plain text content.
- `--max-completion-tokens-cap=20` + request `max_tokens: 1000` → response has `completion_tokens <= 20`.
- Manual smoke test for each of the three panic modes in fix #3 (object without `properties`, array without `items`, property with missing `type`) — container stays up, all requests return HTTP 200.

**End-to-end, real cluster:**
Deployed a locally-built image (linux/amd64) of the v0.9.0 codebase + these commits onto a k8s cluster fronted by an Anthropic-compatible gateway serving Claude Code traffic. With `--skip-tool-validation --tool-choice-override none --max-completion-tokens-cap 50 --max-model-len 131072` set via metadata, all previously-observed failure modes disappear:
- No more `Tool validation failed: jsonschema: … additionalProperties not allowed`
- No more `tool parameters of type <nil>|[null object]|[null array] are not supported`
- No more `interface conversion: interface {} is nil, not map[string]interface {}` panics
- Every response is text; no fabricated tool calls
- Response length bounded; no "infinite response" complaints

## Commits (in order)

| Commit | Title |
|---|---|
| `3d4d862` | `feat: add --skip-tool-validation to bypass the built-in tool schema validator` |
| `08c7c43` | `feat: normalize JSON Schema type in the arg generator (nullable + missing)` |
| `7fe81a4` | `feat: add --tool-choice-override + defensively handle empty properties/items in arg generator` |
| `1562b96` | `feat: decompress gzipped request/response bodies before logging (--log-http)` |
| `8d245c9` | `feat: add --max-completion-tokens-cap to prevent unbounded random-mode responses` |

## Environment / motivation

Deployed sim standing in for `gemma-4` in an airgapped edge k8s box, gateway translating Anthropic → OpenAI on `/v1/messages` → engine on `/v1/chat/completions`. Client: Claude Code (real tool suite). Before these fixes, the sim's first request from Claude Code returned 400 (tool validation), fixing that surfaced tool-generator errors, fixing those surfaced panics, and fixing those surfaced runaway generation. This PR is the composed end state where every one of those failure modes is addressed at its root and each fix is independently useful.

Happy to squash / rebase / split further based on your preferences.

### DIFF
--- a/pkg/common/config.go
+++ b/pkg/common/config.go
@@ -206,6 +206,28 @@ type Configuration struct {
 	// in an object in a tool call, optional, defaults to 50
 	ObjectToolCallNotRequiredParamProbability int `yaml:"object-tool-call-not-required-field-probability" json:"object-tool-call-not-required-field-probability"`
 
+	// SkipToolValidation, when true, bypasses the built-in JSON Schema
+	// validation of incoming tool definitions and forwards them to the
+	// tool-call generator as-is. Matches real vLLM's behavior — real vLLM
+	// does not meta-validate the input tool spec. Useful when clients
+	// (e.g. Anthropic → OpenAI translators, Claude Code) send tool schemas
+	// with fields outside the sim's strict inline schema (default, format,
+	// title, oneOf, $ref, examples, minLength, minimum, pattern, ...).
+	// Default false — backward compat.
+	SkipToolValidation bool `yaml:"skip-tool-validation" json:"skip-tool-validation"`
+
+	// ToolChoiceOverride, when non-empty, ignores the request's tool_choice
+	// and forces the sim to behave as if the client had sent this value
+	// instead. Valid values: "" (default, no override), "none", "auto",
+	// "required". Ignored for a client-forced-specific-function tool_choice.
+	ToolChoiceOverride string `yaml:"tool-choice-override" json:"tool-choice-override"`
+
+	// MaxCompletionTokensCap, when > 0, is a HARD upper bound the sim applies
+	// to the effective completion-token count regardless of the request's
+	// max_tokens / max_completion_tokens. Guards against unbounded random-
+	// mode responses. Default 0 = no cap.
+	MaxCompletionTokensCap int `yaml:"max-completion-tokens-cap" json:"max-completion-tokens-cap"`
+
 	// EnableKVCache defines if kv cache feature will be enabled
 	EnableKVCache bool `yaml:"enable-kvcache" json:"enable-kvcache"`
 	//  KVCacheSize is the maximum number of token blocks in kv cache, the default value is 1024
--- a/pkg/common/parser.go
+++ b/pkg/common/parser.go
@@ -148,6 +148,10 @@ func ParseCommandParamsAndLoadConfig() (*Configuration, error) {
 	f.IntVar(&config.ToolCallNotRequiredParamProbability, "tool-call-not-required-param-probability", config.ToolCallNotRequiredParamProbability, "Probability to add a parameter, that is not required, in a tool call")
 	f.IntVar(&config.ObjectToolCallNotRequiredParamProbability, "object-tool-call-not-required-field-probability", config.ObjectToolCallNotRequiredParamProbability, "Probability to add a field, that is not required, in an object in a tool call")
 
+	f.BoolVar(&config.SkipToolValidation, "skip-tool-validation", config.SkipToolValidation, "Skip JSON Schema validation of incoming tool definitions; forward to the tool-call generator as-is (matches real vLLM behavior)")
+	f.StringVar(&config.ToolChoiceOverride, "tool-choice-override", config.ToolChoiceOverride, "Override the request's tool_choice: \"none\" always returns text (no tool calls), \"auto\" lets the sim decide, \"required\" always calls at least one tool. Empty (default) honors the client's tool_choice.")
+	f.IntVar(&config.MaxCompletionTokensCap, "max-completion-tokens-cap", config.MaxCompletionTokensCap, "Hard upper bound on the effective completion-token count regardless of the request's max_tokens; guards against unbounded random-mode responses. 0 (default) = no cap.")
+
 	f.BoolVar(&config.EnableKVCache, "enable-kvcache", config.EnableKVCache, "Defines if KV cache feature is enabled")
 	f.IntVar(&config.KVCacheSize, "kv-cache-size", config.KVCacheSize, "Maximum number of token blocks in kv cache")
 	f.Float64Var(&config.GlobalCacheHitThreshold, "global-cache-hit-threshold", 0, "Default cache hit threshold [0, 1] for all requests. If a request specifies cache_hit_threshold, it takes precedence")
--- a/pkg/communication/http.go
+++ b/pkg/communication/http.go
@@ -18,6 +18,8 @@ package communication
 
 import (
 	"bufio"
+	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -977,6 +979,35 @@ func truncateBodyForLog(body []byte) string {
 	return string(body[:maxHTTPLogBodyBytes]) + fmt.Sprintf(" ... [truncated, total %d bytes]", len(body))
 }
 
+// decompressForLog returns the body decoded per the given Content-Encoding.
+// Only gzip is handled today (the sole encoding fasthttp emits by default);
+// unknown / empty encodings return the body unchanged. A decompression error
+// falls back to the raw bytes so logging never becomes lossy on best-effort
+// human readability. Result feeds truncateBodyForLog so the maxHTTPLogBodyBytes
+// cap still applies to the DECODED payload (a small gzipped blob can expand
+// a lot). Only mutates a copy for logging — the wire body is never touched.
+func decompressForLog(body []byte, encoding string) []byte {
+	if len(body) == 0 {
+		return body
+	}
+	switch strings.ToLower(strings.TrimSpace(encoding)) {
+	case "", "identity":
+		return body
+	case "gzip":
+		r, err := gzip.NewReader(bytes.NewReader(body))
+		if err != nil {
+			return body
+		}
+		defer r.Close()
+		out, err := io.ReadAll(io.LimitReader(r, maxHTTPLogBodyBytes+1))
+		if err != nil {
+			return body
+		}
+		return out
+	}
+	return body
+}
+
 func (c *Communication) logHTTPMiddleware(next fasthttp.RequestHandler) fasthttp.RequestHandler {
 	return func(ctx *fasthttp.RequestCtx) {
 		c.logHTTPRequest(ctx)
@@ -986,12 +1017,13 @@ func (c *Communication) logHTTPMiddleware(next fasthttp.RequestHandler) fasthttp
 }
 
 func (c *Communication) logHTTPRequest(ctx *fasthttp.RequestCtx) {
+	reqBody := decompressForLog(ctx.Request.Body(), string(ctx.Request.Header.ContentEncoding()))
 	c.logger.V(logging.INFO).Info("HTTP request",
 		"method", string(ctx.Method()),
 		"requestURI", string(ctx.RequestURI()),
 		"remoteAddr", ctx.RemoteAddr().String(),
 		"headers", formatRequestHeaders(&ctx.Request.Header),
-		"body", truncateBodyForLog(ctx.Request.Body()),
+		"body", truncateBodyForLog(reqBody),
 	)
 }
 
@@ -1005,10 +1037,11 @@ func (c *Communication) logHTTPResponse(ctx *fasthttp.RequestCtx) {
 		)
 		return
 	}
+	respBody := decompressForLog(resp.Body(), string(resp.Header.ContentEncoding()))
 	c.logger.V(logging.INFO).Info("HTTP response",
 		"statusCode", resp.StatusCode(),
 		"headers", formatResponseHeaders(&resp.Header),
-		"body", truncateBodyForLog(resp.Body()),
+		"body", truncateBodyForLog(respBody),
 	)
 }
 
--- a/pkg/dataset/dataset.go
+++ b/pkg/dataset/dataset.go
@@ -120,6 +120,12 @@ type DefaultDataset struct {
 	random             *common.Random
 	histogramHelper    *histogramHelper
 	tokenizedResponses []api.Tokenized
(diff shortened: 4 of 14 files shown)


rishi-anand:
Follow-up commit **08c7c43** — `feat: normalize JSON Schema type in the arg generator (nullable + missing)`.

While testing the first commit in a live cluster (`--skip-tool-validation` on, real Claude Code traffic through an Anthropic→OpenAI translator), the sim's random tool-call generator surfaced a second class of errors that also stem from schemas real vLLM tolerates but the sim doesn't:

```
err="tool parameters of type %!s(<nil>) are not supported"
err="tool parameters of type [null object] are not supported"
err="tool parameters of type [null array] are not supported"
```

Three real-world shapes hit the generator's `default` case in `createArgument`:

1. **`"type"` absent** — property with no declared type.
2. **`"type": ["null","object"]`** — the JSON Schema draft-04+ nullable pattern.
3. **`"type": ["null","array"]`** — same pattern for arrays.

The new commit adds a `resolvePrimitiveType()` normalization:
- `[]any` → first non-`"null"` entry wins (useful value for a nullable field), fall back to `"null"` if every entry is `"null"`
- `nil` → `"string"` (broadest safe default when the schema is under-specified)
- other → `""`, which lets the existing "not supported" error surface honestly instead of a silent panic

Plus a `case "null"` that returns Go `nil` so a schema strictly requiring the JSON `null` value produces exactly that.

With both commits applied, the sim now serves Claude Code's tool traffic end-to-end. Happy to squash if that fits the project's history preferences.

rishi-anand:
Follow-up commit **7fe81a4** — `feat: add --tool-choice-override + defensively handle empty properties/items in arg generator`.

Two additions, both driven by testing against real Claude Code / MCP traffic in a deployed sim:

### 1. `--tool-choice-override=[none|auto|required]`

The sim has no model, so its "should I call a tool?" decision in the default `auto` case is effectively a dice roll. With Claude Code attaching its full tool suite (~30+ tools) on every request, the dice roll almost always picks a tool call — the sim ends up fabricating tool invocations for plain conversational queries ("hello" → `playwright.Go back(...)`). Real vLLM doesn't have this issue because a real model decides based on intent.

This adds a deployment-level override (settable via `--config` YAML, so ConfigMap-friendly) that ignores the request's `tool_choice` and forces the sim to behave as if the client sent the override instead:

| Value | Behavior |
|---|---|
| `""` (default) | Honor client's `tool_choice`. Unchanged. |
| `"none"` | Never generate tool calls; always return text. Ideal for text-only testing when the client can't easily be reconfigured per-request. |
| `"auto"` | Force auto (sim decides), even if client sent `required`. |
| `"required"` | Force required, even if client sent `none` / `auto`. |

Verified locally: with `--tool-choice-override=none`, a request with `"tool_choice": "required"` and tools declared returns `finish_reason: stop`, `tool_calls: null` — pure text.

### 2. Nil-safe `items` / `properties` in `createArgument`

The recursive arg generator did unchecked type assertions on `propertyMap["items"]` and `propertyMap["properties"]`, which panic when those fields are absent. Both shapes turn up in real tool schemas:

- Claude Code's `WebFetch` declares `{"type": "object"}` with no `properties`.
- Arrays are commonly declared without an `items` schema.

Now uses comma-ok assertions and returns an empty `{}` / `[]` — a valid sample value in both cases, matching how real vLLM tolerates under-specified schemas by forwarding them to the model instead of synthesizing.

Together with the earlier commits, this branch takes the sim from "400s or panics on nearly every request from Claude Code / MCP clients" to "serves the pipeline end-to-end and lets the operator control the tool-vs-text dial without touching the client."

Three commits on the branch — happy to squash if that fits your history preferences.

mayabar:
Thanks for filing #607 and for the super detailed PR.
This PR is broader than what #607 asks for.
It bundles the `--skip-tool-validation` flag together with several other fixes (nullable/missing type normalization, nil-safe properties/items, --tool-choice-override, --max-completion-tokens-cap, and gzip decompression in --log-http). Each of these is genuinely useful, but reviewing and rolling them back independently is much harder when they land as one PR.
We’re going to merge #613, which addresses #607 exactly and nothing more.
For the remaining changes, could you open a separate issue for each item, especially the panic on missing `properties/items` and the type normalization, which are clearly bugs, and then a follow-up PR per issue? 
That way each change gets its own review thread and can move at its own pace. 

Thanks again — I really appreciate how much detail you put into this.

mayabar:
Closing this PR — the individual changes have been split into their own issues: #636, #637, #638

Thanks again for the detailed write-up here; it made the split much easier.

PR #620, opened 2026-08-12
implement vllm:request_time_per_output_token_seconds metrics

yamt:
## What does this PR do?

implement vllm:request_time_per_output_token_seconds metrics

## Why is this change needed?

vllm compatibility

## How was this tested?

## Checklist

### DIFF
--- a/pkg/llm-d-inference-sim/simulator.go
+++ b/pkg/llm-d-inference-sim/simulator.go
@@ -445,6 +445,7 @@ func (s *VllmSimulator) simulateResponseProcessing(respCtx ResponseContext) {
 			&ResponseInfo{RespCtx: respCtx, Status: ResponseStatusCreated, ChoiceIdx: choiceIdx},
 			s.Context.logger)
 
+		nTokens := 0
 		startDecode := time.Now()
 		if respIsEmpty(respCtx) {
 			common.WriteToChannel(reqCtx.responseChannel(),
@@ -454,6 +455,7 @@ func (s *VllmSimulator) simulateResponseProcessing(respCtx ResponseContext) {
 				for i, token := range respCtx.responseTokens().Tokens {
 					if i != 0 {
 						s.Context.simulateInterTokenLatency()
+						nTokens++
 					}
 
 					tokens := &api.Tokenized{
@@ -478,6 +480,7 @@ func (s *VllmSimulator) simulateResponseProcessing(respCtx ResponseContext) {
 					for i, token := range args.Tokens {
 						if i != 0 {
 							s.Context.simulateInterTokenLatency()
+							nTokens++
 						}
 						respInfo := ResponseInfo{
 							Tokens:    &api.Tokenized{Tokens: []uint32{token}, Strings: []string{args.Strings[i]}},
@@ -493,7 +496,13 @@ func (s *VllmSimulator) simulateResponseProcessing(respCtx ResponseContext) {
 				}
 			}
 		}
-		common.WriteToChannel(s.Context.metrics.reqDecodeTimeChan, time.Since(startDecode).Seconds(), s.Context.logger)
+		decodeTime := time.Since(startDecode).Seconds()
+		meanTPOT := 0.0
+		if nTokens > 0 {
+			meanTPOT = decodeTime / float64(nTokens)
+		}
+		common.WriteToChannel(s.Context.metrics.reqTpotChan, meanTPOT, s.Context.logger)
+		common.WriteToChannel(s.Context.metrics.reqDecodeTimeChan, decodeTime, s.Context.logger)
 
 		if reqCtx.request().SendImage() {
 			s.Context.simulateImageGenerationLatency()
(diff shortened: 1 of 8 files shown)


pkg/llm-d-inference-sim/simulator.go
```diff
@@ -493,7 +493,14 @@ func (s *VllmSimulator) simulateResponseProcessing(respCtx ResponseContext) {
 				}
 			}
 		}
-		common.WriteToChannel(s.Context.metrics.reqDecodeTimeChan, time.Since(startDecode).Seconds(), s.Context.logger)
+		decodeTime := time.Since(startDecode).Seconds()
+		meanTPOT := 0.0
+		respTokens := respCtx.responseTokens()
+		if respTokens != nil && len(respTokens.Tokens) > 1 {
+			meanTPOT = decodeTime / float64((len(respTokens.Tokens) - 1))
```
irar2: This misses the tokens generated for tool calls

irar2 (commented):
All the metrics should also be added to FakeMetrics.

Please add tests.

pkg/llm-d-inference-sim/simulator.go
yamt: fixed

yamt:
> All the metrics should also be added to FakeMetrics.
> 
> Please add tests.

done.

pkg/tests/grpc_metrics_test.go
```diff
@@ -212,6 +212,10 @@ var _ = Describe("gRPC Metrics", Ordered, func() {
 						// buckets higher than 0.75 should be greater than 0, we don't know the exact value since it depends on the random response length
 						count := findIntMetric(metricsLines, getFloatBucketMetricPrefix(common.TestModelName, metricName, boundary))
 						Expect(count).ToNot(BeNil())
+						// note: request TPOT is actually measured. we can't assert the exact timing.
+						if metricName == vllmsim.ReqTPOTMetricName {
+							continue
+						}
```
irar2, line 218: I don't understand why we can't expect the same `>0` for `>0.75`  as for the other two metrics. Isn't it actually the same thing, just a different calculation?

pkg/tests/metrics_test.go
```diff
@@ -590,6 +590,10 @@ var _ = Describe("Simulator metrics", Ordered, func() {
 						// buckets higher than 0.75 should be greater than 0, we don't know the exact value since it depends on the random response length
 						count := findIntMetric(metricsLines, getFloatBucketMetricPrefix(common.TestModelName, metricName, boundary))
 						Expect(count).ToNot(BeNil())
+						// note: request TPOT is actually measured. we can't assert the exact timing.
+						if metricName == vllmsim.ReqTPOTMetricName {
+							continue
+						}
```
irar2, line 596: Likewise

irar2 (changes requested):

pkg/tests/grpc_metrics_test.go
yamt: the existing two metrics just reports the caluclated value:
https://github.com/llm-d/llm-d-inference-sim/blob/1e901d4adf62be070391d1f6c78ca8f886b26457/pkg/llm-d-inference-sim/context.go#L278-L285

the new one measures the actual latency, which is typically a bit longer the calculated value.

irar2: OK, but is the delta so big that 1.0 bucket is empty?

yamt: it's rare (sleep 0.075 + some other logic) takes <= 0.075. i've observed the bucket empty.

irar2: I am sorry, I don't understand. (I was also confused by the incorrect comment about 0.75 instead of 0.075).

inter-token-latency is set to 0.075, then we check that the buckets <0.075 are empty, and the buckets >=0.075, i..e. starting from 0.1 are not empty.  Are you saying that you saw the 0.1 bucket empty? If so, let's allow this bucket to be empty for the new metric, but assert >0 for the next buckets.

Also, could you please fix the comment (0.75 -> 0.075)?

yamt: this test asserts:
```
vllm:request_time_per_output_token_seconds_bucket{model_name="test-model",le="0.025"}
vllm:request_time_per_output_token_seconds_bucket{model_name="test-model",le="0.05"}
  ^
  |  empty
------
  |  not empty
  v
vllm:request_time_per_output_token_seconds_bucket{model_name="test-model",le="0.075"}
vllm:request_time_per_output_token_seconds_bucket{model_name="test-model",le="0.1"}
```
and i observed `vllm:request_time_per_output_token_seconds_bucket{model_name="test-model",le="0.075"}` empty.

anyway, i changed the code to skip only one bucket.

also, note that metrics_test.go and grpc_metrics_test.go uses different inter-token-latency. it's quite confusing too.

> Also, could you please fix the comment (0.75 -> 0.075)?

done.

irar2: Right, very confusing, I will look into this later, this is unrelated to this PR.
Thanks.

irar2 (approved):

PR #624, opened 2026-08-17
Include serving port in ZMQ kv-events topic

tonibofarull:
## What does this PR do?

Builds the ZMQ KV-event topic as `kv@<ip>:<port>@<model>` by including the
simulator's serving port, instead of the previous `kv@<ip>@<model>`. The port is
taken from the per-rank configuration (the port each simulator instance serves
on), so each data-parallel rank publishes on its own topic.

## Why is this change needed?

Bug fix for cache-aware (prefix-cache) routing between `llm-d-inference-sim` and
the EPP prefix-cache scorer:

- The simulator emitted KV-cache events on topic `kv@<ip>@<model>`, keyed by the
  pod IP from `POD_IP`.
- The EPP prefix-cache scorer looks up pods by `<Address>:<Port>`
  ([llm-d-router](https://github.com/llm-d/llm-d-router/blob/dd65342ee538ab76d4b9f7e1efb5d8e95385d06e/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go#L379)).
- The mismatch (`<ip>` vs `<ip>:<port>`) meant the block index never matched, so
  the scorer reported no cache hits and cache-aware routing was effectively
  disabled.

Including the serving port makes the topic match the scorer's addressing, so the
block index can find the pod that owns a cached block. This also keeps the
simulator in sync with real vLLM, which configures the same `<ip>:<port>` topic
via `--kv-events-config`.

For data-parallel deployments, the serving port is the port each rank's
simulator instance serves on:

- `--data-parallel-size N` (all ranks in one process): ranks 1..N-1 serve on
  `--port+1`..`--port+N-1`, so each gets a distinct `<ip>:<port>` topic under
  the shared pod IP.
- `--data-parallel-rank R` (one process per rank, each its own pod): each rank
  serves on its own `--port` and has a distinct `POD_IP`, so topics already
  differ and the serving port needs no offset.

## How was this tested?

- [X] Unit tests added/updated

#### Logs without fix applied

Prepare the dev environment from dd72f043, using the llm-d-router dev environment from https://github.com/llm-d/llm-d-router/pull/2426:

```sh
# In llm-d-inference-sim
$ git checkout dd72f043
$ SIM_TAG=dd72f043 make image-build
# In llm-d-router
$ VLLM_SIMULATOR_TAG=dd72f043 VLLM_REPLICA_COUNT_D=2 KV_CACHE_ENABLED=true make env-dev-kind
```

Send the prompt:

```sh
$ PREFIX="Prefix to make the prompt large enough (16 tokens by default) to force hit cache in next attempt."
$ curl -s http://localhost:30080/v1/completions -d "{\"model\":\"TinyLlama/TinyLlama-1.1B-Chat-v1.0\",\"prompt\":\"${PREFIX} Say hello.\",\"max_tokens\":10,\"temperature\":0}" | jq
```

See logs

```sh
$ kubectl logs -f deployments/tinyllama-1-1b-chat-v1-0-endpoint-picker
{"level":"debug","ts":1787003362.3830872,"caller":"scheduling/scheduler_profile.go:217","msg":"Calculated score","x-request-id":"e5c3fbb1-af88-4242-a59a-8ac0ba920172","objectiveKey":"","incomingModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","targetModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","priority":0,"plugin":"prefix-cache-scorer/prefix-cache-scorer","endpoint":{"name":"vllm-d-c54d97cc4-rnpbb-rank-0","namespace":"default"},"score":0}
{"level":"debug","ts":1787003362.3830957,"caller":"scheduling/scheduler_profile.go:217","msg":"Calculated score","x-request-id":"e5c3fbb1-af88-4242-a59a-8ac0ba920172","objectiveKey":"","incomingModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","targetModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","priority":0,"plugin":"prefix-cache-scorer/prefix-cache-scorer","endpoint":{"name":"vllm-d-c54d97cc4-hgkl4-rank-0","namespace":"default"},"score":0}
```

Even if we repeat the prompt, the scores remain 0.

#### Logs with fix applied

```sh
# In llm-d-inference-sim
$ git checkout b10e054c
$ SIM_TAG=dev make image-build
# In llm-d-router
$ VLLM_SIMULATOR_TAG=dev VLLM_REPLICA_COUNT_D=2 KV_CACHE_ENABLED=true make env-dev-kind
```

Send the prompt twice:

```sh
$ PREFIX="Prefix to make the prompt large enough (16 tokens by default) to force hit cache in next attempt."
$ curl -s http://localhost:30080/v1/completions -d "{\"model\":\"TinyLlama/TinyLlama-1.1B-Chat-v1.0\",\"prompt\":\"${PREFIX} Say hello.\",\"max_tokens\":10,\"temperature\":0}" | jq
$ curl -s http://localhost:30080/v1/completions -d "{\"model\":\"TinyLlama/TinyLlama-1.1B-Chat-v1.0\",\"prompt\":\"${PREFIX} Say hello.\",\"max_tokens\":10,\"temperature\":0}" | jq
```

See logs

```sh
$ kubectl logs -f deployments/tinyllama-1-1b-chat-v1-0-endpoint-picker
{"level":"debug","ts":1787071620.7134564,"caller":"scheduling/scheduler_profile.go:217","msg":"Calculated score","x-request-id":"a491b43b-53cb-4b9d-ab8e-cd82fcd7b682","objectiveKey":"","incomingModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","targetModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","priority":0,"plugin":"prefix-cache-scorer/prefix-cache-scorer","endpoint":{"name":"vllm-d-999bcd8d5-j2kq5-rank-0","namespace":"default"},"score":0}
{"level":"debug","ts":1787071620.713466,"caller":"scheduling/scheduler_profile.go:217","msg":"Calculated score","x-request-id":"a491b43b-53cb-4b9d-ab8e-cd82fcd7b682","objectiveKey":"","incomingModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","targetModelName":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","priority":0,"plugin":"prefix-cache-scorer/prefix-cache-scorer","endpoint":{"name":"vllm-d-999bcd8d5-97gnh-rank-0","namespace":"default"},"score":1}
```

After the second request, the pod that cached the prefix scores 1 while the other stays 0, so the scorer routes to it.

## Checklist

- [X] Commits are signed off (`git commit -s`) per [DCO](https://github.com/llm-d/.github/blob/main/PR_SIGNOFF.md)
- [X] Code follows project [contributing guidelines](https://github.com/llm-d/.github/blob/main/CONTRIBUTING.md)
- [X] Tests pass locally (`make test`)
- [X] Linters pass (`make lint`)
- [X] Documentation updated (if applicable)

### DIFF
--- a/docs/kv-cache.md
+++ b/docs/kv-cache.md
@@ -104,12 +104,14 @@ When `enable-kvcache` is true and `zmq-endpoint` is set, the simulator publishes
 ### Topic format
 
 ```
-kv@<POD_IP>@<model-name>
+kv@<POD_IP>:<serving-port>@<model-name>
 ```
 
-Example: `kv@10.0.0.1@Qwen/Qwen2.5-1.5B-Instruct`
+Example: `kv@10.0.0.1:8000@Qwen/Qwen2.5-1.5B-Instruct`
 
 The model name used is the base model name (`--model`), not a served-model-name alias.
+
+The `<ip>:<port>` pair matches how the EPP prefix-cache scorer addresses pods (as `<Address>:<Port>`), so the block index can look up the pod that owns a cached block. The serving port is the port each simulator instance serves on, so each data-parallel rank publishes on its own topic.
 
 ### Message format
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -198,7 +198,7 @@ In addition, as we are using klog, the following parameters are available:
 - `VLLM_SERVER_DEV_MODE`: when set to `1`, enables vLLM development mode. Currently used as an additional gate for the `/sleep` endpoint: even with `--enable-sleep-mode`, `/sleep` is a no-op unless `VLLM_SERVER_DEV_MODE=1` is set in the simulator's environment.
 - `POD_NAME`: the simulator pod name. If defined, the response will contain the HTTP header `x-inference-pod` with this value, and the HTTP header `x-inference-port` with the port that the request was received on 
 - `POD_NAMESPACE`: the simulator pod namespace. If defined, the response will contain the HTTP header `x-inference-namespace` with this value
-- `POD_IP`: the simulator pod IP address. Used in kv-events topic name.
+- `POD_IP`: the simulator pod IP address. Used together with the serving port (`--port`) to build the kv-events topic `kv@<ip>:<port>@<model>` (see [KV cache](kv-cache.md#topic-format)).
 Example of definition in yaml: 
   ```yaml
   env:
--- a/manifests/zmq-listener/README.md
+++ b/manifests/zmq-listener/README.md
@@ -86,7 +86,7 @@ The vLLM deployment configures event publishing via:
   \"enable_kv_cache_events\":true,
   \"publisher\":\"zmq\",
   \"endpoint\":\"tcp://zmq-listener-service:5557\",
-  \"topic\":\"kv@${POD_IP}@Qwen/Qwen3-0.6B\"
+  \"topic\":\"kv@${POD_IP}:8000@Qwen/Qwen3-0.6B\"
 }"
 ```
 
--- a/manifests/zmq-listener/deploy_vllm.yaml
+++ b/manifests/zmq-listener/deploy_vllm.yaml
@@ -28,7 +28,7 @@ spec:
               --block-size 16 \
               --prefix-caching-hash-algo sha256_cbor \
               --kv-transfer-config '{"kv_connector":"NixlConnector", "kv_role":"kv_both"}' \
-              --kv-events-config "{\"enable_kv_cache_events\":true,\"publisher\":\"zmq\",\"endpoint\":\"tcp://zmq-listener-service:5557\",\"topic\":\"kv@${POD_IP}@Qwen/Qwen3-0.6B\"}"
+              --kv-events-config "{\"enable_kv_cache_events\":true,\"publisher\":\"zmq\",\"endpoint\":\"tcp://zmq-listener-service:5557\",\"topic\":\"kv@${POD_IP}:8000@Qwen/Qwen3-0.6B\"}"
           env:
           - name: NAMESPACE
             valueFrom:
--- a/pkg/kv-cache/block_cache.go
+++ b/pkg/kv-cache/block_cache.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -86,7 +87,7 @@ func newBlockCache(ctx context.Context, config *common.Configuration, logger log
 		}
 	}
 
-	topic := CreateKVEventsTopic(config.IP, config.Model)
+	topic := CreateKVEventsTopic(config.IP, config.Port, config.Model)
 
 	var replayer *kvEventsReplayer
 	if config.KVEventsReplayEndpoint != "" {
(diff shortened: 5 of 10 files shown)


docs/kv-cache.md
```diff
@@ -111,6 +111,16 @@ Example: `kv@10.0.0.1@Qwen/Qwen2.5-1.5B-Instruct`
 
 The model name used is the base model name (`--model`), not a served-model-name alias.
 
+When `KV_EVENTS_INCLUDE_VLLM_PORT=true` is set, the serving port (`--port`, default `8000`) is appended to `POD_IP`, so the topic becomes:
+
+```
+kv@<POD_IP>:<port>@<model-name>
```
yamt: isn't it simpler to introduce, say, `POD_ID`, and use kv@<POD_ID>@<model-name>?

tonibofarull: According to the documentation it requires the port,

https://github.com/llm-d/llm-d/blob/v0.9.0/guides/precise-prefix-cache-routing/README.md#how-it-works

As well as in the implementation side,

https://github.com/llm-d/llm-d-router/blob/dd65342ee538ab76d4b9f7e1efb5d8e95385d06e/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go#L379

yamt: my point is it's simpler to use a single environment variable (POD_IDENTIFIER) than two. (POD_IP, KV_EVENTS_INCLUDE_VLLM_PORT)

tonibofarull: I wanted to preserve the current behavior. Since the port is already known by the app, I think separating the opt-in for including it avoids requiring users to provide information they might not know or could get wrong.

mayabar: @yamt It would be problematic to define a single port through an environment variable when the data-parallelism size is greater than 1.

In that case, multiple simulator instances are created, each with its own port and KV cache. Therefore, each KV-cache instance should use its corresponding port, as defined in the configuration.

mayabar:
@yamt 
Suggestion: instead of adding a new environment variable, let's just change the current behavior. The original design was intended to stay in sync with llm-d, and let's keep it that way.

As a next step, we could add support for `--kv-events-config` (real vLLM's command line option) to replace the current command-line option - introduce the new flag, mark `--zmq-endpoint` as deprecated, and remove it after two releases. Let's handle that in a separate PR to keep this one focused.

pkg/common/config.go
```diff
@@ -365,8 +367,12 @@ type LoraModule struct {
 }
 
 func newConfig() *Configuration {
+	ip := os.Getenv(podIPEnv)
+	if ip != "" && os.Getenv(kvEventsIncludeVLLMPort) == trueString {
+		ip += ":" + strconv.Itoa(vLLMDefaultPort)
```
mayabar: Default port (vLLMDefaultPort) is used here while the real one, defined in the configuration should be used. 
In addition, please pay attention to DP support when data parallelism size is > 1. In this case port should be taken from the simulator instance and not the command line

tonibofarull:
@mayabar, @yamt, I have changed the code to set the port by default. Please review :bow: 
Notice that I test with https://github.com/llm-d/llm-d-router/pull/2426

yamt (approved):
lgtm

mayabar (approved):

PR #631, opened 2026-08-19
fix: prevent silent token/request drops from full or oversized channels

irar2:
- `newRequests` full now returns 429 instead of silently dropping the request
- `respChan` was undersized for multiple choices (and text completions with multiple prompts),  now sized off actual sub-request count
- Added `--max-n-choices`, requests over the limit get a 400.
- `sendNonStream`/`sendStream` now guard against a nil choice, returning a clean 500 instead of panicking
- Metrics channels now sized off running vs. waiting request counts separately instead of one shared sum
- Includes tests for all of the above and doc updates for the new flag


mayabar:
Do we really need `--max-n-choices`?

mayabar (approved):

PR #639, opened 2026-08-20
fix: tolerate tool schemas that omit properties or items

### LINKED ISSUE #636: bug: panic in tool-call arg generator when tool schema omits `properties` or `items`
## Summary

`createArgument` in [`pkg/llm-d-inference-sim/tools_utils.go`](pkg/llm-d-inference-sim/tools_utils.go) does unchecked type assertions on `propertyMap["items"]` (line 299) and `propertyMap["properties"]` (line 323). When a tool schema legitimately omits either field, the request panics.

## Reproduction

Any tool declaration where a property has:

- `"type": "object"` without `properties`, e.g. Claude Code's `WebFetch` tool.
- `"type": "array"` without `items`.

Sending such a schema and forcing a tool call (e.g. `tool_choice: "required"`) panics:

```
interface conversion: interface {} is nil, not map[string]interface {}
```

## Expected behavior

Real vLLM tolerates under-specified schemas because it forwards them to the model instead of synthesizing arguments. The sim should tolerate them by treating the missing fields as empty:

- Missing `properties` -> empty object `{}`.
- Missing `items` -> empty array `[]` (or skip element generation).

## Proposed fix

Use comma-ok assertions at the two sites and return an empty value when the field is absent. Prior art: commit `7fe81a4` from PR #608 already implements this shape.

## Context

Split from PR #608, which bundled this bug fix with other changes (https://github.com/llm-d/llm-d-inference-sim/pull/608#issuecomment-5231416954).

himanshu1573:
## What does this PR do?

Makes the tool-call argument generator tolerate schemas that declare an array or object without describing its contents. `createArgument` asserted `propertyMap["items"]` and `propertyMap["properties"]` unconditionally; both are now comma-ok, yielding an empty array and an empty object respectively.

## Why is this change needed?

Both assertions panic when the field is absent:

```
interface conversion: interface {} is nil, not map[string]interface {}
```

`"type": "object"` with no `properties` and `"type": "array"` with no `items` are valid JSON Schema, and real vLLM accepts them because it forwards the schema to the model instead of synthesizing arguments. The sim crashes the request instead, which blocks load tests against representative tool suites.

## How was this tested?

- [ ] Manual testing performed

New unit tests cover the generator's handling of under-specified schemas: an array property with no `items` produces an empty array, an object property with no `properties` produces an empty object, and an object that does declare `properties` still generates its nested values. Each test was confirmed to panic at the two call sites before the fix.

## Checklist

`make presubmit` and `make test` both pass locally: all 8 suites green. No user-facing behavior or flags change, so no documentation update applies.

## Related Issues

Fixes #636

### DIFF
--- a/pkg/llm-d-inference-sim/tools_arg_generator_test.go
+++ b/pkg/llm-d-inference-sim/tools_arg_generator_test.go
@@ -1,0 +1,80 @@
+/*
+Copyright 2025 The llm-d-inference-sim Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package llmdinferencesim
+
+import (
+	"encoding/json"
+
+	"github.com/llm-d/llm-d-inference-sim/pkg/common"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("createArgument with under-specified schemas", func() {
+	var (
+		config *common.Configuration
+		random *common.Random
+	)
+
+	BeforeEach(func() {
+		config = &common.Configuration{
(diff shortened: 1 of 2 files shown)


himanshu1573:
@mayabar when you have a moment, could you approve the workflow run? `CI - PR Checks` is sitting at `action_required` since runs on fork PRs need maintainer approval, so lint-and-test has not executed yet.

Locally `make presubmit` and `make test` both pass (8/8 suites). The other checks - DCO, signed-commits, and the three FOSSA jobs - are green.

mayabar (approved):

PR #644, opened 2026-08-23
Rework context-window validation for random and echo modes

irar2:
Random mode now checks only that the prompt leaves room for one response token: `max_tokens` is no longer part of context-window validation, and is clamped instead if it would overflow the window.
Echo mode now requires the prompt to fit twice (it's echoed back as the response) and additionally rejects `max_tokens` smaller than the prompt. 
The PR updates tests and docs accordingly.


pkg/tests/simulator_test.go
```diff
@@ -410,15 +415,41 @@ var _ = Describe("Simulator", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			openaiclient, params := getOpenAIClientAndChatParams(client, model, prompt, false)
+			// would have been rejected under the old prompt+max_tokens<=max-model-len check
+			params.MaxTokens = openai.Int(1000000)
+
+			resp, err := openaiclient.Chat.Completions.New(ctx, params)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.Choices).To(HaveLen(1))
+			Expect(resp.Model).To(Equal(model))
```
mayabar, line 437: we may check that returned text is <= 10 tokens

pkg/tests/simulator_test.go
```diff
@@ -1011,8 +1112,11 @@ var _ = Describe("Simulator", func() {
 
 		It("Should not drop tokens across many concurrent requests (echo mode)", func() {
 			ctx := context.TODO()
+			// echo mode requires max-model-len > 2*prompt tokens (the prompt is echoed
+			// back as the response), so max-model-len is set just above 2*15 to stay
+			// as tight as that constraint allows.
 			args := []string{"cmd", "--model", common.TestModelName, "--mode", common.ModeEcho,
-				"--max-num-seqs", "100", "--max-model-len", "16", "--max-waiting-queue-length", "1"}
+				"--max-num-seqs", "100", "--max-model-len", "31", "--max-waiting-queue-length", "1"}
```
mayabar: why 31 and not 30?

pkg/tests/simulator_test.go
irar2: It's a bit difficult, we don't have tokens in the response.

pkg/tests/simulator_test.go
irar2: Done

mayabar (approved):

PR #649, opened 2026-08-25
Add Engine abstraction selected via --engine

irar2:
Introduces an Engine interface (pkg/engine) that determines how the simulator's
configuration is constructed. The engine is selected via a new --engine flag (or the
engine key in a --config YAML file), resolved before the rest of the configuration is
built. Only one engine exists today, vllm, and it produces exactly the existing
configuration and behavior - this is a seam for future engines, not a behavior change.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -42,9 +42,9 @@ type Engine interface {
 	NewConfiguration() *common.Configuration
 }
 
-// Get returns the Engine identified by name, or an error listing the valid values if
-// name is not recognized.
-func Get(name string) (Engine, error) {
+// New creates the Engine identified by name, or returns an error listing the valid
+// values if name is not recognized.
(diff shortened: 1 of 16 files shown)


pkg/engine/engine.go
```diff
@@ -0,0 +1,76 @@
...
+	// Name returns the engine's identifier, the value accepted by --engine.
+	Name() string
+	// NewConfiguration returns a Configuration populated with this engine's defaults,
+	// before any YAML or command-line overrides are applied.
+	NewConfiguration() *common.Configuration
+}
+
+// Get returns the Engine identified by name, or an error listing the valid values if
+// name is not recognized.
+func Get(name string) (Engine, error) {
```
mayabar: consider renaming to New()

mayabar (approved):

PR #650, opened 2026-08-25
Make tests could pass in restricted networks via leveraging proxy env

### LINKED ISSUE #643: Restricted networks: HF_ENDPOINT/proxies ignored by both the render container and modelExists() — real models silently degrade to the simulated tokenizer
### What happened?

The test/tokenizer plumbing assumes `huggingface.co` is directly reachable, and none of it honors `HF_ENDPOINT` or proxy settings. On a machine where `huggingface.co` is not directly reachable (proxy-only network), this shows up in two places that share the same root cause.

**1) The render container receives no network-related environment, so the `pkg/dataset` suite cannot start at all.**

The suite fails in `[BeforeSuite]` and all 61 specs are skipped:

```
[FAILED] Unexpected error:
    failed to start render container: generic container: start container:
    started hook: wait until ready: context deadline exceeded
Ran 0 of 61 Specs — A BeforeSuite node failed so all tests were skipped.
```

The vLLM render container's own log shows why — it cannot download the tokenizer:

```
ERROR [repo_utils.py:47] Error retrieving file list: [Errno 101] Network is unreachable, retrying 1 of 2
...
pydantic_core._pydantic_core.ValidationError: 1 validation error for ModelConfig
  Value error, Invalid repository ID or local directory specified: 'Qwen/Qwen2-VL-2B-Instruct'.
```

`TokenizerManager.startRenderContainer` (`pkg/tokenizer/test_utils.go`) starts `vllm/vllm-openai-cpu:v0.21.0` via `testcontainers.Run(...)` **without any environment passthrough**. `huggingface_hub` inside the container honors `HF_ENDPOINT` / `HF_TOKEN` / proxy variables, but there is currently no way to supply them from the host environment:

- Host-level `HF_ENDPOINT` (a corporate or public mirror) never reaches the container.
- Docker CLI proxy config (`~/.docker/config.json` → `proxies.default`) does not apply either, because testcontainers-go creates containers through the Docker API, not the CLI.

This makes `make test` unrunnable for contributors behind firewalls/mirrors (a very common setup in some regions and in corporate CI), even though the code under test is unaffected.

**2) `modelExists()` hardcodes `huggingface.co`, so real models silently degrade to the simulated tokenizer.**

After working around (1) locally (env passthrough + a reachable `HF_ENDPOINT`), the suite goes from 0/61 executed to 44 executed / 43 passed — and the one remaining failure turned out to be this second, subtler instance of the same root cause, not a stale fixture:

```
CustomDataset [It] should return correct prompt hash in bytes
  Expected "a963e695d1cde52298f5157bc5eb95dde6331aabd04a46f3c25a24ebfdcabf7d"
  to equal "73205d2e432e6b117e0b75cdddeac019ee863f4b524f75bf57c15c5a47a445e4"
```

The hardcoded fixture hash is actually correct:

- Asking the pinned render server (`vllm/vllm-openai-cpu:v0.21.0`, model `Qwen/Qwen2-VL-2B-Instruct`, tokenizer revision `895c3a49`) directly:
  `POST /v1/completions/render {"model":"Qwen/Qwen2-VL-2B-Instruct","prompt":"1, 2, 3, 4, 5, 6, 7, 8, 9, 10"}`
  returns `token_ids = [16,11,220,17,11,220,18,11,220,19,11,220,20,11,220,21,11,220,22,11,220,23,11,220,24,11,220,16,15]`, and `sha256(LE-uint32(tokens))` is exactly the expected `73205d2e…`.
- Instrumenting the failing spec shows what the suite actually hashed:
  `tokens=[873244444 2312110321 923577301 2312110321 …]`, `strings=["1", ", ", "2", ", ", …]`
  — FNV-hash pseudo-tokens from `SimpleTokenizer`, whose hash is the observed `a963e695…`.

The mechanism is in `pkg/tokenizer/tokenizer.go`:
```go
func modelExists(model string) bool {
	url := "https://huggingface.co/api/models/" + model

	statusCode, _, err := fasthttp.Get(nil, url)
	if err != nil {
		return false
	}

	return statusCode == fasthttp.StatusOK
}
```

The URL is hardcoded (ignores `HF_ENDPOINT`), and `fasthttp.Get` does not use `HTTPS_PROXY` either — so in a restricted network it returns `false` for every model, and both call sites treat `false` as "dummy model" and **silently** fall back to `SimpleTokenizer`: the test-suite selection in `TokenizerManager.newTokenizer()` (`pkg/tokenizer/test_utils.go:162`) and, more importantly, the production selection in `tokenizer.New()` (`tokenizer.go:62`).

That means a simulator deployed in a restricted network with a genuine model quietly runs on regex/FNV pseudo-tokens even when its render server is up and reachable — anything depending on real token IDs (KV-cache block hashing, prefix-cache-aware routing experiments, dataset prompt hashes) silently behaves differently, with only an INFO-level log hinting at it.

### How to reproduce it (as minimally and precisely as possible)

Part 1 — render container:

1. Block direct access to `huggingface.co` (or run in any network where it is unreachable).
2. `export HF_ENDPOINT=xxxx` (a reachable mirror).
3. `go test ./pkg/dataset/...` → BeforeSuite fails as above; the exported `HF_ENDPOINT` is silently ignored.

Part 2 — modelExists fallback (with Part 1 worked around so the render container starts):

1. Same restricted network, render container able to fetch the tokenizer via the mirror.
2. `go test ./pkg/dataset/ -args -ginkgo.focus='should return correct prompt hash in bytes'`
3. The spec fails with `a963e695…` vs expected `73205d2e…`, identically when run in isolation (so not an ordering/mutation effect). Logging `validDB[0].tokenizedInput.Tokens` in the spec shows FNV pseudo-tokens instead of Qwen token IDs.

### Anything else we need to know?

Proposed fix, in two parts matching the two manifestations:

1. **Render container**: propagate a small allowlist of network-related variables (`HF_ENDPOINT`, `HF_TOKEN`, `HTTP(S)_PROXY`/`NO_PROXY` upper/lower case) from the host into the container via `testcontainers.WithEnv`. Verified locally: with this and a reachable `HF_ENDPOINT`, the suite goes from 0/61 executed to 44 executed / 43 passed. One caveat worth documenting in README "Testing locally": proxy values must be reachable *from inside the container* — a host proxy on `127.0.0.1` must not be forwarded as-is; exporting only `HF_ENDPOINT` is the simplest working setup.
2. **`modelExists`**: honor `HF_ENDPOINT` (default `https://huggingface.co`), and consider `net/http` (which honors `HTTPS_PROXY` via `http.ProxyFromEnvironment`) instead of `fasthttp.Get` for this one probe. Additionally: log the fallback at WARN (not INFO) when a real-looking repo id degrades to the simulated tokenizer — the silent degradation is the expensive part; and in the test suite, don't degrade at all — `TokenizerManager.Init` already starts a real render container for the Qwen model, so a failed existence probe should be an error there, since the dataset fixtures are only meaningful against real tokenization.

With both fixed, the full dataset suite passes in restricted networks. Happy to send a PR (DCO signed) for any subset once the preferred direction is agreed.

### Environment

- main @ eec055a (2026-08-20)
- macOS (Apple Silicon) + Colima, Docker 29.7.2, Go 1.27

Myasuka:
## What does this PR do?

 Two commits, both scoped to the tokenizer path:

  1. **`tokenizer/test_utils`: forward HF and proxy env into the render container.**
     `startRenderContainer` now passes an allowlist of HuggingFace and proxy-related host env (`HF_ENDPOINT`, `HF_TOKEN`, `HF_HUB_ENDPOINT`, and both cases of `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY`) into the vLLM render container via `testcontainers.WithEnv`.
  2. **`tokenizer`: select tokenizer by `--render-url` instead of probing `huggingface.co`.** Remove `modelExists()` and its `fasthttp` probe. The simulator now uses `HFTokenizer` when `--render-url` is set, and `SimpleTokenizer` otherwise. When `--render-url` is missing but the model name looks like a HuggingFace repo id (contains `/`), the fallback is logged at WARN so restricted-network deployments do not silently serve pseudo-token ids. 

## Why is this change needed?

On restricted networks (no direct access to `huggingface.co`), the previous
  tokenizer path failed silently in two ways:

  1. The test render container tried to reach `huggingface.co` directly to download `Qwen/Qwen2-VL-2B-Instruct`, so the `pkg/dataset` `BeforeSuite` failed and skipped all 61 specs.
  2. In production, `modelExists()` used `fasthttp.Get` against a hardcoded `https://huggingface.co/api/models/<model>` URL. `fasthttp` does not honor `HTTPS_PROXY` and the URL cannot be pointed at a mirror, so the probe returned `false` for every real model and the caller fell back to `SimpleTokenizer`.

 The selection rule follows the discussion on the issue: real render service is opt-in via `--render-url`; a real-looking repo id without a render URL is a loud WARN, not a silent degradation.

## How was this tested?

## Checklist

- [x Code follows project [contributing guidelines](https://github.com/llm-d/.github/blob/main/CONTRIBUTING.md)

## Related Issues

Fixes #643

### DIFF
--- a/pkg/tokenizer/test_utils.go
+++ b/pkg/tokenizer/test_utils.go
@@ -19,6 +19,7 @@ package tokenizer
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -29,6 +30,29 @@ import (
 	testcontainers "github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
 )
+
+// hostNetworkEnvKeys are host environment variables forwarded into the render
+// container so HuggingFace mirrors and proxies configured on the host reach
+// the tokenizer download inside the container.
+var hostNetworkEnvKeys = []string{
+	"HF_ENDPOINT",
+	"HF_TOKEN",
+	"HF_HUB_ENDPOINT",
+	"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY",
+	"http_proxy", "https_proxy", "no_proxy",
+}
+
+// collectHostNetworkEnv returns the subset of hostNetworkEnvKeys that are set
+// on the host, suitable for testcontainers.WithEnv.
+func collectHostNetworkEnv() map[string]string {
+	env := map[string]string{}
+	for _, key := range hostNetworkEnvKeys {
+		if v := os.Getenv(key); v != "" {
+			env[key] = v
+		}
+	}
+	return env
+}
 
 type TokenizerManager struct {
 	testTokenizer Tokenizer
@@ -75,14 +99,14 @@ func (tm *TokenizerManager) Init(ctx context.Context, logger logr.Logger) error 
 	// var err error
 	// create tokenizer for Qwen model
 	// Use longer timeout (30s) for render requests as the container may need time to fully initialize
-	tm.qwenTokenizer, err = tm.newTokenizer(ctx, logger, renderURL, common.QwenModelName, 30*time.Second, 60*time.Second)
+	tm.qwenTokenizer, err = tm.newHFTokenizer(ctx, logger, renderURL, common.QwenModelName, 30*time.Second, 60*time.Second)
 	if err != nil {
 		cleanup()
 		return err
 	}
 
 	// create tokenizer for multimodal model
-	tm.mmTokenizer, err = tm.newTokenizer(ctx, logger, renderURL, common.QwenModelName, 30*time.Second, 60*time.Second)
+	tm.mmTokenizer, err = tm.newHFTokenizer(ctx, logger, renderURL, common.QwenModelName, 30*time.Second, 60*time.Second)
 	if err != nil {
 		cleanup()
 		return err
@@ -106,6 +130,7 @@ func (tm *TokenizerManager) startRenderContainer(ctx context.Context, model stri
 		testcontainers.WithEntrypoint("vllm"),
 		testcontainers.WithCmd("launch", "render", model, "--port=8000"),
 		testcontainers.WithTmpfs(map[string]string{"/.cache": "rw"}),
+		testcontainers.WithEnv(collectHostNetworkEnv()),
 		testcontainers.WithWaitStrategy(
 			wait.ForHTTP("/health").
 				WithPort("8000/tcp").
@@ -140,17 +165,6 @@ func (tm *TokenizerManager) startRenderContainer(ctx context.Context, model stri
 	}
 
 	return address, cleanup, nil
-}
-
-func (tm *TokenizerManager) newTokenizer(ctx context.Context, logger logr.Logger, renderURL, model string,
-	timeout, mmTimeout time.Duration) (Tokenizer, error) {
-	if modelExists(model) {
-		// for real model create HF tokenizer
-		return tm.newHFTokenizer(ctx, logger, renderURL, model, timeout, mmTimeout)
-	} else {
-		// for dummy model create simple tokenizer
-		return tm.newSimpleTokenizer(model), nil
-	}
 }
 
 // create Simple Tokenizer
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ To run the vLLM simulator in a standalone test environment with a real model:
    ./bin/llm-d-inference-sim --model Qwen/Qwen2.5-0.5B-Instruct --port 8000 --render-url http://localhost:8082
    ```
 
-**Note:** If the model is not a real HuggingFace model, there is no need to run the vLLM render server.
+**Note:** If you do not pass `--render-url`, the simulator uses the simulated tokenizer and no vLLM render server is required.
 
 ## Testing locally
 ### Prerequisites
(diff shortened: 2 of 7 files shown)


pkg/tokenizer/test_utils.go
```diff
@@ -144,13 +169,11 @@ func (tm *TokenizerManager) startRenderContainer(ctx context.Context, model stri
...
-	} else {
-		// for dummy model create simple tokenizer
-		return tm.newSimpleTokenizer(model), nil
-	}
+	// The manager has already started a render container for this model. Any
+	// tokenizer failure here should surface as an error rather than silently
+	// fall back to the simulated tokenizer, which would produce pseudo-token
+	// ids and break dataset assertions built against the real tokenizer.
+	return tm.newHFTokenizer(ctx, logger, renderURL, model, timeout, mmTimeout)
 }
```
mayabar: newTokenizer function is not required any more, there is `newHFTokenizer` which is similar to already used `newSimpleTokenizer`

mayabar (commented):
@Myasuka Thanks for the PR, it looks good, just a small comment

pkg/tokenizer/test_utils.go
```diff
@@ -30,6 +31,29 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
```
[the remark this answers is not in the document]
Myasuka: I think import `os` module would not introduce too much overhead.

pkg/tokenizer/test_utils.go
Myasuka: Thanks for the review, already addressed.

PR #653, opened 2026-08-26
fix: decode gzip-compressed bodies before logging them with --log-http

### LINKED ISSUE #637: bug: --log-http writes gzip-compressed body bytes to the log
## Summary

`logHTTPRequest` / `logHTTPResponse` in [`pkg/communication/http.go`](pkg/communication/http.go#L1006-L1031) log `ctx.Request.Body()` and `resp.Body()` as-is. When fasthttp has applied gzip content-encoding (which it does automatically for responses when the client sends `Accept-Encoding: gzip`), the log entry contains the compressed bytes rather than readable content.

## Reproduction

Run the sim with `--log-http`. Send a request from a Prometheus scraper or any client that advertises `Accept-Encoding: gzip`. Response body in `kubectl logs` is unreadable binary rather than the JSON payload.

Symmetric problem on the request side for clients that send gzip-encoded request bodies.

## Expected behavior

The log entry should show the decoded body. Raw wire bytes stay unchanged; only the log output is decompressed.

## Proposed fix

Introduce a small helper (e.g. `decompressForLog`) that inspects the relevant `Content-Encoding` header and gunzips before handing the bytes to `truncateBodyForLog`. Applied symmetrically at the two call sites. Prior art: commit `1562b96` from PR #608.

## Context

Split from PR #608, which bundled this fix with other unrelated changes.

zaid-fakhruddin:
## What does this PR do?

Decodes bodies sent with `Content-Encoding: gzip` before `--log-http` logs them. Previously, `logHTTPRequest`/`logHTTPResponse` fed `Body()` straight to `truncateBodyForLog`. Both call sites now go through `bodyForLog`, which decompresses at most `maxHTTPLogBodyBytes+1` bytes, so truncation applies to the decoded output. If decoding fails, the raw bytes are logged, as before. Wire bytes unchanged.

fasthttp's `BodyUncompressedWithLimit` is unsuitable here: past the limit it returns `(nil, ErrBodyTooLarge)` and drops the partial decode, which re-creates the bug for large bodies. This uses the same bounded read as commit 1562b96; when a body decodes past the cap, the truncation marker reports the read bound, not the full decoded size.

## Why is this change needed?

promhttp gzip-compresses `/metrics` when gzip is negotiated via `Accept-Encoding`, so the response body in the log is binary. Same on the request side for clients that gzip bodies.

## How was this tested?

Five specs added in `pkg/communication/http_log_test.go`, the first three against the same JSON payload:

- unencoded, it is logged as is
- gzipped with `Content-Encoding: gzip`, it decodes back to the payload
- when the body is labeled gzip but is not actually gzipped, the raw bytes are logged, confirming the fallback really runs
- nil or empty, an empty string is logged
- a body that decodes past `maxHTTPLogBodyBytes` keeps the decoded prefix and gets the truncation marker

Tested manually with `--log-http --v 2`: `/metrics` responses under `Accept-Encoding: gzip` and gzipped POST bodies (including one that decodes to 600 KB) are logged as text, while the bytes on the wire stay gzipped. `make presubmit` and `make test` pass.

## Checklist

## Related Issues

Fixes #637


irar2:
@zaid-fakhruddin Looks good. The only thing I would add is a test for empty/nil payload.

pkg/communication/http_log_test.go
```diff
@@ -0,0 +1,52 @@
...
+	})
+
+	It("truncates the decoded body at the limit", func() {
+		body := fasthttp.AppendGzipBytes(nil, bytes.Repeat([]byte("a"), maxHTTPLogBodyBytes+1))
+
+		got := bodyForLog(body, []byte("gzip"))
+		Expect(got).To(HavePrefix("aaa"))
+		Expect(got).To(HaveSuffix(fmt.Sprintf("[truncated, total %d bytes]", maxHTTPLogBodyBytes+1)))
+	})
+})
```
irar2, line 57: Please add tests for empty/nil payload

irar2 (changes requested):

pkg/communication/http_log_test.go
zaid-fakhruddin: added, thanks!

irar2 (approved):

PR #655, opened 2026-08-27
Move engine-specific routing and gRPC registration out of communication

irar2:
Moves vLLM-specific wiring out of `pkg/communication`'s hardcoded startup path and into `pkg/engine`, via two new `Engine` methods:
- `Routes(c *communication.Communication) []communication.Route` - vLLM specific endpoints  now come from `vllmEngine.Routes`, bound in `communication`'s `startHTTPServer` via a plain loop instead of being hardcoded there.
- `GRPC() communication.GRPCRegistrar` - vLLM's gRPC service registration also moves to the engine. A nil registrar means the gRPC listener isn't started at all.

`pkg/simulator` no longer imports or stores an `Engine`

### DIFF
--- a/pkg/engine/vllm.go
+++ b/pkg/engine/vllm.go
@@ -16,7 +16,14 @@ limitations under the License.
 
 package engine
 
-import "github.com/llm-d/llm-d-inference-sim/pkg/common"
+import (
+	"github.com/valyala/fasthttp"
+	"google.golang.org/grpc"
+
+	"github.com/llm-d/llm-d-inference-sim/pkg/common"
+	"github.com/llm-d/llm-d-inference-sim/pkg/communication"
+	"github.com/llm-d/llm-d-inference-sim/pkg/communication/grpc/pb"
+)
 
 type vllmEngine struct{}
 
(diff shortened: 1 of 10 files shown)


pkg/engine/vllm.go
```diff
@@ -29,3 +36,31 @@ func (vllmEngine) NewConfiguration() *common.Configuration {
...
+// Routes returns vLLM's HTTP-specific endpoints: the render debug endpoints, the
+// Responses and Messages APIs, the vLLM-specific generate endpoint, LoRA adapter
+// load/unload, the Mooncake bootstrap query, and the sleep/wake_up/is_sleeping
+// dev-mode endpoints.
+func (vllmEngine) Routes(c *communication.Communication) []communication.Route {
+	return []communication.Route{
+		{Method: fasthttp.MethodPost, Path: "/v1/chat/completions/render", Handler: c.HandleChatCompletionsRender},
+		{Method: fasthttp.MethodPost, Path: "/v1/completions/render", Handler: c.HandleTextCompletionsRender},
+		{Method: fasthttp.MethodPost, Path: "/v1/responses", Handler: c.HandleResponses},
+		{Method: fasthttp.MethodPost, Path: "/v1/messages", Handler: c.HandleMessages},
```
mayabar, line 51: maybe add a comment here that this is Antropic's api, but for now it's here in vllm
same for /query

mayabar (approved):

PR #659, opened 2026-08-30
Add audio content

### LINKED ISSUE #658: Support audio and video multimedia payloads in /v1/chat/completions
## Summary

Please add support in `llm-d-inference-sim` for multimedia (audio and video) payloads in `/v1/chat/completions` requests, matching the multi-modal content format used by real vLLM serving of models such as `Qwen/Qwen2.5-Omni-3B`.

Today the simulator handles text and images chat completions. To benchmark and validate coordinator/routing paths that carry multi-modal payloads end-to-end (large `audio_url` / `video_url` content parts, prefill/decode split behavior, etc.) without needing GPUs and the real Omni model, the simulator needs to at least accept and parse these content parts without erroring out.

## Requested behavior

- Accept `messages[].content` as an array of typed parts, including:
  - `{"type": "text", "text": "..."}`
  - `{"type": "audio_url", "audio_url": {"url": "..."}}`
  - `{"type": "video_url", "video_url": {"url": "..."}}`
  - (Ideally also `image_url`, if not already supported, for parity.)
- Parse and account for these parts in whatever token/latency simulation model is in use (even a rough size-based approximation would be enough for benchmarking).
- Continue to return a valid OpenAI-compatible chat completion / stream response.

### Audio

```json
{
  "model": "Qwen/Qwen2.5-Omni-3B",
  "stream": true,
  "max_tokens": 128,
  "messages": [
    {
      "role": "user",
      "content": [
        {"type": "text", "text": "What is being said in this audio?"},
        {"type": "audio_url", "audio_url": {"url": "https://raw.githubusercontent.com/QwenLM/Qwen2-Audio/main/assets/audio/1272-128104-0000.flac"}}
      ]
    }
  ]
}
```

### Video

```json
{
  "model": "Qwen/Qwen2.5-Omni-3B",
  "stream": true,
  "max_tokens": 128,
  "messages": [
    {
      "role": "user",
      "content": [
        {"type": "text", "text": "Describe what happens in this video."},
        {"type": "video_url", "video_url": {"url": "https://samplelib.com/mp4/sample-5s.mp4"}}
      ]
    }
  ]
}
```

## Motivation

We are running coordinator/router performance benchmarks against `llm-d-inference-sim` as a GPU-free stand-in for vLLM. Multimedia scenarios (audio, video) are part of our benchmark matrix, and today those requests cannot be exercised against the simulator.

mayabar:
Adds support for audio_url content type in `/chat/completions` and `chat/completions/render`

Fixes #658

### DIFF
--- a/pkg/tokenizer/tokenizer.go
+++ b/pkg/tokenizer/tokenizer.go
@@ -141,7 +141,7 @@ func (st *SimpleTokenizer) RenderMessages(messages []api.Message) ([]uint32, []s
 }
 
 // stubMMFeaturesForMessages synthesizes per-modality mm_hashes and placeholders
-// for image_url, input_audio, and video_url blocks; nil if none present.
+// for image_url, input_audio, audio_url, and video_url blocks; nil if none present.
 func stubMMFeaturesForMessages(messages []api.Message, totalTokens int) *api.RenderMMFeatures {
 	type item struct {
 		modality, prefix, identifier string
@@ -149,7 +149,7 @@ func stubMMFeaturesForMessages(messages []api.Message, totalTokens int) *api.Ren
 	}
 
 	var items []item
-	var imgIdx, audIdx, inputAudioIdx, vidIdx int
+	var imgIdx, audIdx, vidIdx int
 	for _, msg := range messages {
 		for _, block := range msg.Content.Structured {
 			switch block.Type {
@@ -163,8 +163,8 @@ func stubMMFeaturesForMessages(messages []api.Message, totalTokens int) *api.Ren
 				if block.InputAudio == nil || block.InputAudio.Data == "" {
 					continue
 				}
-				items = append(items, item{mmModalityAudio, "input_audio", block.InputAudio.Data, inputAudioIdx})
-				inputAudioIdx++
+				items = append(items, item{mmModalityAudio, "input_audio", block.InputAudio.Data, audIdx})
+				audIdx++
 			case "audio_url":
 				if block.AudioURL == nil || block.AudioURL.Url == "" {
 					continue
--- a/pkg/tokenizer/tokenizer.go
+++ b/pkg/tokenizer/tokenizer.go
@@ -149,7 +149,7 @@ func stubMMFeaturesForMessages(messages []api.Message, totalTokens int) *api.Ren
 	}
 
 	var items []item
-	var imgIdx, audIdx, vidIdx int
+	var imgIdx, audIdx, inputAudioIdx, vidIdx int
 	for _, msg := range messages {
 		for _, block := range msg.Content.Structured {
 			switch block.Type {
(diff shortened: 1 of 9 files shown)


dmitripikus:
Several comments:

1. IsMultiModal missed audio_url — [pkg/api/render.go:82](vscode-webview://10jnhlbkpeqplhv757jvbc4o9bjra45ojsua8quvcvp9pr2shb6p/pkg/api/render.go#L82)
Detects only image_url. Audio-only requests get mm=false, so the render client uses the short timeout instead of mmTimeout. Users setting --mm-render-timeout see it silently ignored for audio-only (and video-only) payloads. Tokenizer side now handles audio_url but this side doesn't — the two paths are inconsistent.

2. PlainText doesn't render audio_url/input_audio/video_url — [pkg/api/response.go:197](vscode-webview://10jnhlbkpeqplhv757jvbc4o9bjra45ojsua8quvcvp9pr2shb6p/pkg/api/response.go#L197)
Only image_url gets a text stub. Audio-only messages tokenize as just the role prefix, so SimpleTokenizer's token count collapses and stubMMFeaturesForMessages's span math (totalTokens/len(items)) degenerates to a single placeholder at offset 0. Prefix-cache/routing hashing for audio isn't representative of payload size.

3. New audio_url test doesn't actually cover audio_url — [pkg/tests/chat_completions_test.go:442](vscode-webview://10jnhlbkpeqplhv757jvbc4o9bjra45ojsua8quvcvp9pr2shb6p/pkg/tests/chat_completions_test.go#L442)
The request carries a text block and the audio block, and asserts only prompt_tokens > 0. The text alone satisfies that — dropping audio_url handling entirely wouldn't fail this test. Either remove the text block or assert on MMHashes["audio"].

4. Shared audio bucket, independent index counters — [pkg/tokenizer/tokenizer.go:172](vscode-webview://10jnhlbkpeqplhv757jvbc4o9bjra45ojsua8quvcvp9pr2shb6p/pkg/tokenizer/tokenizer.go#L172)
input_audio uses inputAudioIdx, audio_url uses audIdx, but both go into mmHashes["audio"]. With both present, position 0 appears twice in the same bucket. Any consumer that treats the bucket index as "position within the audio modality" gets misaligned placeholders.

5. Stale docstring — [pkg/tokenizer/tokenizer.go:144](vscode-webview://10jnhlbkpeqplhv757jvbc4o9bjra45ojsua8quvcvp9pr2shb6p/pkg/tokenizer/tokenizer.go#L144)
Comment still says "for image_url, input_audio, and video_url" — needs audio_url added.

pkg/tokenizer/tokenizer.go
```diff
@@ -163,7 +163,13 @@ func stubMMFeaturesForMessages(messages []api.Message, totalTokens int) *api.Ren
 				if block.InputAudio == nil || block.InputAudio.Data == "" {
 					continue
 				}
-				items = append(items, item{mmModalityAudio, "audio", block.InputAudio.Data, audIdx})
+				items = append(items, item{mmModalityAudio, "input_audio", block.InputAudio.Data, audIdx})
```
irar2: Why not leave it "audio" for backward compatibility?

irar2 (approved):

PR #670, opened 2026-09-01
Move model-validation and image/mooncake decisions from communication into SimContext

irar2:
`ValidateBaseModel` was misplaced on `Simulator` despite touching only `SimContext` state, and `pkg/communication`'s HTTP layer held two more simulator decisions inline: the Omni image-emission roll and the per-rank Mooncake engine-id map. This PR moves all three onto `SimContext` and exposes them through `endpoint.Runtime`, continuing the same decoupling pattern used for sleep/wake-up (#669).

### DIFF
--- a/pkg/communication/http.go
+++ b/pkg/communication/http.go
@@ -184,7 +184,7 @@ func (c *Communication) handleRender(req endpoint.RenderableRequest, respBuilder
 		c.sendError(ctx, err, false)
 		return
 	}
-	if err := c.simulator.ValidateBaseModel(req.GetModel()); err != nil {
+	if err := c.runtime.ValidateBaseModel(req.GetModel()); err != nil {
 		c.sendError(ctx, err, false)
 		return
 	}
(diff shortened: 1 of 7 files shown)


pkg/communication/http.go
mayabar, line 256: Maybe this logic should also be inside the runtime? This is simulation logic not communication.
Same for CacheThresholdFinishReasonHeader.
@irar2 What do you think?

pkg/endpoint/runtime_test.go
```diff
@@ -47,7 +47,12 @@ func (f *fakeRuntime) GetResponseTokens(req api.Request) (*api.Tokenized, string
...
-func (f *fakeRuntime) IsSleeping() bool                     { return false }
+func (f *fakeRuntime) KVCacheOnRequestEnd(requestID string)      {}
+func (f *fakeRuntime) Sleep() bool                               { return false }
+func (f *fakeRuntime) WakeUp(activateKVCache bool)               {}
+func (f *fakeRuntime) IsSleeping() bool                          { return false }
+func (f *fakeRuntime) ValidateBaseModel(model string) *api.Error { return nil }
+func (f *fakeRuntime) ShouldSendImage(headerOverride bool) bool  { return headerOverride }
+func (f *fakeRuntime) MooncakeEngineMap() map[string]map[string]string {
+	return nil
+}
```
mayabar: nit: maybe use same one-line formatting as above?

pkg/communication/http.go
irar2: This code reads the header (needs to be here), converts it (can be moved, but kind of feels like the conversion should be here too, where the value is read), and then sends the error (should be here). I don't think there is much too move here...

pkg/endpoint/runtime_test.go
irar2: done

mayabar (approved):

PR #678, opened 2026-09-02
feat: add derender endpoints, the inverse of the render endpoints

### LINKED ISSUE #677: Support vLLM's derender endpoints (/v1/chat/completions/derender, /v1/completions/derender)
### What would you like to be added?

Add support for vLLM's derender endpoints, the inverse of the render endpoints:

- `POST /v1/chat/completions/derender` — accepts `{model?, generate_response, prompt_tokens?, chat_request?}` and returns a `/v1/chat/completions` response
- `POST /v1/completions/derender` — accepts `{model?, generate_responses[], prompt_tokens[]?, completion_request?}` and returns a `/v1/completions` response

Both detokenize the raw `token_ids` carried in the generate response(s) and assemble the client-facing OpenAI response: `id` taken verbatim from `request_id`, `finish_reason` passthrough, `usage` computed from `prompt_tokens` plus output token counts, `kv_transfer_params` passthrough, and (for completions) a flat running choice index across responses — matching vLLM's behavior and error semantics (404 unknown model; 400 for `stream: true`, missing/empty generate responses, empty `token_ids`, `prompt_tokens` length mismatch).

Detokenization per tokenizer mode:

- **Simulated tokenizer:** token ids are one-way FNV hashes, so the tokenizer keeps an in-memory reverse mapping of every id it has produced (render, generation, tokenization). Ids produced by the same instance round-trip to their original text; unknown ids render as `<unk_ID>` placeholders.
- **HuggingFace tokenizer:** decoding is forwarded to the upstream render service's `/v1/completions/derender` at `--render-url`.

### Why is this needed?

The simulator already implements `/render` and `/inference/v1/generate`, i.e. two of the three legs of vLLM's disaggregated serving flow. vLLM recently added the postprocessing leg as derender endpoints (RFC: vllm-project/vllm#42729, shipped under `vllm/entrypoints/scale_out/`). Without them, llm-d disaggregated flows (render → generate → derender) cannot be exercised end to end against the simulator.

## Out of scope (deferred, protocol allows adding later without breaking changes)

- Streaming derender (`stream: true` per-chunk variants) — rejected with `400 Bad Request`
- Tool-call and reasoning parsing from the decoded token stream (`chat_request`/`completion_request` are accepted and ignored)
- Logprobs `token_id:N` placeholder resolution (the simulator's generate endpoint does not emit logprobs)

shimib:
## What does this PR do?

Adds vLLM-compatible derender endpoints, completing the disaggregated flow next to the existing `/render` and `/inference/v1/generate` support:

- `POST /v1/chat/completions/derender` — `{model?, generate_response, prompt_tokens?, chat_request?}` → `/v1/chat/completions` response
- `POST /v1/completions/derender` — `{model?, generate_responses[], prompt_tokens[]?, completion_request?}` → `/v1/completions` response

Both detokenize the raw `token_ids` in the generate response(s) and assemble the OpenAI response with vLLM's semantics: `id` verbatim from `request_id`, `finish_reason` and `kv_transfer_params` passthrough (dropped with a warning when completions responses disagree), `usage` from `prompt_tokens` plus output token counts, flat running choice index for completions, `model` defaulting to the served model. Errors match vLLM: 404 for an unknown model; 400 for `stream: true`, a missing/empty generate response, empty `token_ids`, or a `prompt_tokens` length mismatch.

The `Tokenizer` interface gains `Detokenize`:

- **Simulated tokenizer:** keeps an in-memory reverse map of the one-way FNV token-id hashes (populated on every encode), so `render → generate → derender` round-trips the actual text within an instance; unknown ids render as `<unk_ID>`.
- **HuggingFace tokenizer:** forwards decoding to the render service's `/v1/completions/derender` at `--render-url`.

Out of scope (deferred; the protocol carries the original-request context so these can be added without breaking changes): streaming derender, tool-call/reasoning parsing, logprobs `token_id:N` placeholder resolution.

## Why is this change needed?

vLLM shipped derender as the postprocessing leg of disaggregated serving (RFC: vllm-project/vllm#42729). Without it, llm-d disaggregated flows cannot be exercised end to end against the simulator.

## How was this tested?

New coverage: request validation and response assembly for both endpoints (usage math, kv passthrough/drop, flat indexing, placeholder fallback); simple-tokenizer detokenization round-trip and unknown-id handling; integration specs for simulate and proxy modes plus every error case. Manual verification against the built binary: chat and completions round-trips, the full `render → generate → derender` flow recovering generated text from raw token ids, and all error paths.

Note: the proxy-mode integration specs skip with an explicit message when the render container's vLLM does not serve `/derender` — the current CI image (`vllm/vllm-openai-cpu:v0.21.0`) may predate it; bumping the image would exercise the proxy path for real.

## Checklist

- [ ] Tests pass locally (`make test`) — container-backed suites cannot run on my machine (no Docker); all container-less suites pass, including with `-race`, and `make lint` is clean

## Related Issues

Fixes #677


docs/http-endpoints.md
```diff
@@ -464,4 +486,15 @@ Pre-tokenized prompts on `/v1/completions/render` (a token-id array, or an array
...
+
+Both endpoints are synchronous and stateless. `usage` is computed from the request's `prompt_tokens` (0 when omitted) plus the total number of token IDs across all choices. Tool-call and reasoning parsing is not performed; the decoded text is returned as plain `content`. Streaming derender (`stream: true`) is not supported and is rejected with `400 Bad Request`.
+
+Detokenization depends on the active tokenizer (selected automatically based on `--model`):
+
+- **HuggingFace tokenizer** (real model): token IDs are decoded through the upstream vLLM render service's `/v1/completions/derender` endpoint at `--render-url`. The upstream must be a vLLM version that serves the derender endpoints.
+- **Simulated tokenizer** (dummy model): token IDs are pseudo-hashes, so the simulator keeps an in-memory reverse mapping of every ID it has produced (through render, generation, or tokenization) and decodes by lookup. Only IDs produced by the same simulator instance round-trip to their original text; unknown IDs are rendered as `<unk_ID>` placeholders.
```
mayabar: Please add explanation how it will work with a dataset (when using `dataset-path` or `dataset-url`)

docs/http-endpoints.md
```diff
@@ -464,4 +486,15 @@ Pre-tokenized prompts on `/v1/completions/render` (a token-id array, or an array
...
+- **Simulated tokenizer** (dummy model): the simulator tokenizes locally using its regex-based splitter. For chat requests containing `image_url`, `audio_url`, `input_audio`, or `video_url` blocks, synthetic `mm_features` are produced so multimodal-aware downstream code paths can be exercised without a real renderer.
+
+### Derender endpoints
+
+`/v1/completions/derender` and `/v1/chat/completions/derender` mirror vLLM's `/derender` behavior — the inverse of the render endpoints. They accept generation results carrying raw token IDs (the shape produced by `/inference/v1/generate`) and return a client-facing OpenAI response, letting disaggregated flows (`render` → `generate` → `derender`) be exercised end to end.
+
+Both endpoints are synchronous and stateless. `usage` is computed from the request's `prompt_tokens` (0 when omitted) plus the total number of token IDs across all choices. Tool-call and reasoning parsing is not performed; the decoded text is returned as plain `content`. Streaming derender (`stream: true`) is not supported and is rejected with `400 Bad Request`.
+
+Detokenization depends on the active tokenizer (selected automatically based on `--model`):
```
mayabar: Please pay attention that the current tokenizer depends on `--render-url` parameter.

pkg/tokenizer/render_client.go
```diff
@@ -80,6 +80,24 @@ func (rc *renderClient) render(endpoint string, payload []byte, mm bool) ([]uint
...
+
+	var resp api.TextCompletionsResponse
+	if err := json.Unmarshal(respBody, &resp); err != nil {
+		return "", fmt.Errorf("unmarshal derender response: %w", err)
+	}
+	if len(resp.Choices) == 0 {
+		return "", errors.New("derender returned no choices")
+	}
+	return resp.Choices[0].Text, nil
+}
```
mayabar: missing support for /v1/chat/completions/derender - it should contains message with role and content

pkg/communication/http.go
```diff
@@ -208,6 +211,54 @@ func (c *Communication) handleRender(req endpoint.RenderableRequest, respBuilder
...
+		c.sendError(ctx, &errToSend, false)
+		return
+	}
+	if err := req.ValidateBody(); err != nil {
+		c.sendError(ctx, err, false)
+		return
+	}
+	// an empty model means the served model, matching vLLM
+	if model := req.GetModel(); model != "" {
+		if err := c.runtime.ValidateBaseModel(model); err != nil {
```
mayabar: In case of error in this function it will be "The model `%s` is a LoRA adapter and is not supported by the **render** endpoints."
Consider passing the endpoint name to the function

pkg/tokenizer/tokenizer.go
```diff
@@ -49,6 +52,11 @@ type baseTokenizer struct {
 
 type SimpleTokenizer struct {
 	baseTokenizer
+
+	// idToString reverses the one-way token-id hashes for Detokenize; it
+	// only contains ids this instance has produced
+	mu         sync.RWMutex
+	idToString map[uint32]string
```
mayabar: Consider to define some limitation on this map size. Maybe using LRU list, or return an error

mayabar (commented):
Thanks for this PR!

The implementation looks good, some small comments below

pkg/communication/http.go
shimib: Good catch, done - ValidateBaseModel now takes the endpoint family name, so the message reports "render" or "derender" according to the caller.

pkg/tokenizer/tokenizer.go
shimib: Done - the reverse map is now bounded at 262,144 entrieswith least-recently-encoded eviction; evicted ids decode as <unk_ID> placeholders, same as unknown ids. I preferred eviction over returning an error so long-running instances keep serving derender and only lose round-trip fidelity for stale ids.

mayabar (approved):

PR #681, opened 2026-09-03
test: shorten gRPC KV-cache metrics check

### LINKED ISSUE #679: Review test  "should send correct kv cache usage metrics via gRPC"
This test uses relatively long TTFT and ITL values and includes a 15-second wait. Please check whether it’s possible to shorten these durations.

0z5a:
## What does this PR do?

Shortens the gRPC KV-cache usage metrics spec without changing its assertions:

- reduces simulated TTFT and inter-token latency from 2s/2s to 500ms/250ms
- replaces fixed 1s, 3s, and 15s sleeps with bounded polling for the running and completed states
- closes each metrics response body after polling

## Why is this change needed?

The test currently spends 19 seconds in fixed sleeps even when the expected state is already observable. Polling preserves both lifecycle assertions while completing as soon as the simulator reaches each state, which makes the test faster and less dependent on fixed timing assumptions.

## How was this tested?

- [x] Unit tests updated
- [x] Manual timing comparison performed

On the same L20 host with Go 1.26.6 and warm module/build caches, the focused spec reported:

| Revision | Result | Spec duration |
| --- | --- | ---: |
| baseline `0153b223` | pass | 19.082s |
| this change | pass | 2.344s |

That is an 87.7% reduction, or about 8.1x faster.

The updated focused spec also passed 10 consecutive isolated runs, ranging from 2.331s to 2.368s. The isolation only bypassed the package-wide Qwen render-container setup and selected the dummy tokenizer for this already-tokenized request; those temporary harness changes are not part of this PR.

Additional checks:

```text
go test ./pkg/tests -run "^$" -count=1
ok github.com/llm-d/llm-d-inference-sim/pkg/tests 0.046s [no tests to run]

go vet ./pkg/tests
```

A normal focused suite run was also attempted. Its package-wide `BeforeSuite` timed out waiting for the unrelated vLLM CPU render container health check, before any specs ran; the focused isolation above was used to validate this test behavior directly.

## Checklist

- [x] Commits are signed off per DCO
- [x] Code follows project contributing guidelines
- [ ] Full `make presubmit` completed
- [x] Targeted `go vet` passes
- [ ] Documentation updated (not applicable)

## Related Issues

Fixes #679

### DIFF
--- a/pkg/tests/grpc_metrics_test.go
+++ b/pkg/tests/grpc_metrics_test.go
@@ -286,7 +286,7 @@ var _ = Describe("gRPC Metrics", Ordered, func() {
 	It("should send correct kv cache usage metrics via gRPC", func() {
 		ctx := context.TODO()
 		args := []string{"cmd", "--model", common.QwenModelName, "--mode", common.ModeEcho,
-			"--time-to-first-token", "2s", "--inter-token-latency", "2s",
+			"--time-to-first-token", "500ms", "--inter-token-latency", "250ms",
 			"--max-num-seqs", "3", "--enable-kvcache", "true", "--kv-cache-size", "16", "--block-size", "8"}
 
 		_, comm, httpClient, err := startServerHandle(ctx, common.ModeEcho, args, map[string]string{"POD_IP": "localhost"})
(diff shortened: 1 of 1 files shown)


pkg/tests/grpc_metrics_test.go
```diff
@@ -286,7 +286,7 @@ var _ = Describe("gRPC Metrics", Ordered, func() {
 	It("should send correct kv cache usage metrics via gRPC", func() {
 		ctx := context.TODO()
 		args := []string{"cmd", "--model", common.QwenModelName, "--mode", common.ModeEcho,
-			"--time-to-first-token", "2s", "--inter-token-latency", "2s",
+			"--time-to-first-token", "500ms", "--inter-token-latency", "250ms",
```
irar2: Maybe even reduce ttft to 300ms and inter-token to 100ms?

irar2 (commented):
@0z5a Using Eventually is a good idea. We will create a separate issue to consider using it in other metrics tests.

pkg/tests/grpc_metrics_test.go
0z5a: Validated the suggested 300ms TTFT and 100ms inter-token latency in local commit `2769097` (not pushed yet), keeping the existing polling, timeouts, and metric assertions.

The focused spec passed 10 consecutive isolated runs (1.011-1.019s). These used a temporary Go overlay to skip the render-container setup and select the dummy tokenizer for the already-tokenized request; the overlay is not part of the commit. The normal focused suite was blocked in BeforeSuite because Docker is unavailable locally. Package compilation and `go vet ./pkg/tests` also passed.

irar2: So, then please push the commit? Or am I missing something?

0z5a: already pushed, sorry for mistake

PR #682, opened 2026-09-03
chore: remove deprecated dummy tokenizer flag

### LINKED ISSUE #652: Remove deprecated --force-dummy-tokenizer
Deprecated in v0.11.0, should be removed in v0.13.0

0z5a:
## What does this PR do?

Removes the deprecated `--force-dummy-tokenizer` option and its configuration field, tokenizer selection branch, tests, and documentation. Tests that only needed the simulated tokenizer now rely on the supported behavior: omit `--render-url`.

A configuration test verifies that the removed CLI option is rejected instead of being silently accepted.

## Why is this change needed?

The option was deprecated in v0.11.0 and scheduled for removal in v0.13.0. Keeping the dead compatibility path increases configuration and test surface after its replacement has been available for multiple releases.

## How was this tested?

Commands run in `golang:1.26.6`:

```text
go test ./pkg/common -count=1
ok github.com/llm-d/llm-d-inference-sim/pkg/common 16.252s

go test ./... -run "^$" -count=1
all packages compiled successfully
```

A full `go test ./... -count=1` was also attempted with Docker access. Non-render packages passed, but the three suites that start the pinned `vllm/vllm-openai-cpu:v0.21.0` renderer timed out in `BeforeSuite` because the test environment could not reach Hugging Face. No test assertion failed.

## Checklist

## Related Issues

Fixes #652


irar2:
Hi @0z5a, it's too early to remove this flag. It should be removed in v0.13.0.

0z5a:
> Hi @0z5a, it's too early to remove this flag. It should be removed in v0.13.0.

i see

Issue #676, opened 2026-09-02
Add "strict" mode for full vLLM request-parity validation


mayabar:
## Summary
Introduce a new strict mode that makes the simulator's request handling behave identically to real vLLM with respect to field validation. When strict mode is enabled, request bodies and headers must fully match vLLM's contract; any deviation returns a `400 Bad Request` with the same error shape vLLM would produce.

## Motivation
Today the simulator accepts a subset of vLLM's request fields and performs limited validation. This is convenient for basic load testing but diverges from real vLLM behavior in ways that hide client bugs and make the simulator unsuitable for verifying request-shaping logic in routers, EPPs, and SDKs. A strict mode gives integration tests a high-fidelity target without requiring a real model server.

## Scope

When strict mode is on:

- **Required fields**: add any required request fields currently missing from the simulator and validate them (including fields that already exist but are not fully validated).
- **Optional fields**: accept the full set of optional fields vLLM accepts, and validate their types, ranges, and mutual constraints.
- **Headers**: validate request headers that vLLM validates (e.g. `content-type`, `auth-related` headers where applicable).
- **Unknown / non-vLLM fields**: mirror vLLM's behavior exactly — reject if vLLM rejects, ignore if vLLM ignores.
- **Error responses**: on any validation failure, return `400 Bad Request` with a body matching vLLM's error format.

When strict mode is off, current lenient behavior is preserved (no regressions for existing users).

## Success criteria
A configuration flag (e.g. `--strict`/ `strict: true`) toggles the mode.
- Validation tests exercise each required field, each optional field, header validation, and unknown-field handling, and assert responses match vLLM byte-for-byte (status code, error code, message shape).
- Tests are driven by fixtures captured from real vLLM where possible, so parity is checkable rather than asserted from memory.
- Endpoints covered: at minimum /v1/completions and /v1/chat/completions; extend to other endpoints as their vLLM contracts are documented.

## Out of scope
- Changing default (non-strict) behavior.
- Simulating vLLM's model-execution errors (only request-validation parity).

PR #683, opened 2026-09-03
feat: add opt-in vLLM sampling validation

0z5a:
## What does this PR do?

Adds opt-in validation for `/v1/chat/completions` and `/v1/completions` with `--strict`. The vLLM engine embeds complete request schemas pinned to vLLM 0.21.0 and runs schema and semantic validation when requests arrive, before tokenization or generation. No user-supplied schema file is required.

The engine owns sampling ranges, sequence limits, streaming/logprob combinations, tool usage, structured-output constraints and repetition-detection checks. Generic request checks remain shared with lenient mode. The request types carry no validation-only state, and `min_tokens` has one validation site.

## Why is this change needed?

The simulator otherwise leaves unused request fields unchecked. Versioning the schema and semantic rules together prevents strict mode from mixing constraints from different vLLM releases. The checked-in audit documents every request-field category and the model-dependent limits that require a real serving configuration; strict mode does not assume a vocabulary size or fixed logprob capacity.

Schema errors use the OpenAI error envelope without promising Pydantic coercion or formatting. Schema defaults affect validation only. Context-limited effective token maxima and structured-output backend compilation are outside the receipt-time validator.

## How was this tested?

- [x] Unit tests added/updated: bundled nested schemas, exact sequence-limit messages, integer boundaries, streaming/logprob combinations, tool and structured-output constraints, repetition detection, omitted/null/explicit token limits, and strict/lenient HTTP dispatch.
- [x] Actual-process HTTP E2E: 35 checks across strict and lenient servers, both completion endpoints, successful JSON/SSE generation, common validation, schema errors, semantic errors and integer overflow.
- [ ] Manual comparison against a live vLLM server for this revision.

Passed targeted tests for communication, vLLM engine, endpoint, common and API; targeted `go vet`; compilation of all packages; and `make lint` with Go 1.26.6. Regenerating the snapshot from the pinned clean vLLM checkout produced identical bytes. The installed `govulncheck` reported no reachable vulnerabilities.

`make presubmit` passed signature/DCO, module tidiness, formatting and lint checks, then failed downloading the latest vulnerability-checker tool due to a network timeout. The installed checker was run separately.

## Checklist

- [x] Commits are signed off (`git commit -s`) per DCO
- [x] Code follows project contributing guidelines
- [x] Documentation updated

## Related Issues

Related to #676


0z5a:
Hi @irar2 Would you also mind having a review on this?

mayabar (commented):
The simulator has so far parsed only the fields that affect response generation.

In strict mode, we should validate the entire request against the OpenAI API specification, along with the vLLM-specific extension fields. This means verifying that all required fields are present and that every supplied field has a valid value and is consistent with related parameters.

The implementation in this PR validates only part of the request. For example:
- `n` has an upper limit configured by `VLLM_MAX_N_SEQUENCES`, but this limit is not currently validated.
- When `max_tokens` is omitted, vLLM applies a default value. However, `min_tokens` is not validated against that effective maximum.

Rather than incrementally adding validation for individual fields, I suggest using an existing library that validates OpenAI-compatible requests, if one is available. Alternatively, we could generate or obtain a JSON Schema from vLLM’s OpenAPI specification:

```bash
curl http://localhost:8000/openapi.json > openapi.json
```
Then we can apply standard JSON Schema validation for the common request structure and add explicit validation for constraints that depend on runtime configuration or relationships between fields, such as `VLLM_MAX_N_SEQUENCES` and the effective `max_tokens` value.

0z5a:
Thanks, I agree that strict mode should use the full request schema. I reworked this to compile the two completion request schemas from a target vLLM OpenAPI 3.1 snapshot, using the JSON Schema library already in the repository. `--strict` now requires `--strict-openapi`; schema validation covers required fields, types, nested structures, and extension fields before the sampling checks.

I also added the `VLLM_MAX_N_SEQUENCES` upper bound and validation of `min_tokens` against the effective maximum: the explicit limit or the schema's omitted-field default, capped by the remaining context window after tokenization. Explicit null and the chat-only `max_completion_tokens` override are handled separately.

The communication, endpoint, common, and API tests, targeted `go vet`, package compilation checks, and lint with Go 1.26.6 passed. The new tests cover schema failures, runtime limits, defaults, context bounds, and lenient behavior. I have not rerun the live vLLM differential benchmark for this revision. OpenAPI does not include all Python/model-dependent validators, so the docs state that limitation and do not promise identical Pydantic error text.

docs/configuration.md
```diff
@@ -27,6 +27,8 @@ Some environment variables (for example `POD_NAME`, `POD_NAMESPACE`) are not ove
 - `enable-sleep-mode`, `no-enable-sleep-mode`: Enable or disable sleep mode feature. When enabled, the simulator can be put to sleep via the `/sleep` endpoint and woken up via the `/wake_up` endpoint
 - `enable-request-id-headers`: Enable including X-Request-Id header in responses. When enabled, the simulator will include the request ID in response headers
 - `log-http`: When true, logs each HTTP request and response at INFO (method, URI, remote address, headers, and body when buffered). Gzip-encoded bodies are decoded before logging. Streamed response bodies (for example SSE) are not logged. Use only in trusted environments; may include secrets such as `Authorization` headers.
+- `strict`: Validate `/v1/chat/completions` and `/v1/completions` against a target vLLM request schema, plus sampling and runtime constraints. Requires `strict-openapi`. Optional, default is false.
+- `strict-openapi`: Path to the target vLLM server's exported OpenAPI 3.1 JSON document. Required only with `strict`. The schemas are compiled once at startup; external schema references are rejected.
```
mayabar: I would consider downloading the OpenAPI schema manually and keeping a copy in the project.

In any case, we will still need to periodically review changes to vLLM’s request validation code.

This would make the simulator easier to use, since users would not need to download the latest schema themselves or ensure that the schema remains in sync with the simulator’s additional internal validation logic.

docs/configuration.md
```diff
@@ -227,3 +229,18 @@ Example of definition in yaml:
...
+Export the schema from the vLLM version you want to simulate:
+
+```sh
+curl --fail http://localhost:8000/openapi.json -o vllm-openapi.json
+llm-d-inference-sim --strict --strict-openapi vllm-openapi.json
+```
+
+Keep the snapshot with the target vLLM version and model configuration. Strict mode checks the entire JSON request against that snapshot, including nested structures and extension fields. Unknown-field acceptance follows the supplied schema. Schema defaults are read for validation; they do not modify response generation.
+
+Set `VLLM_MAX_N_SEQUENCES` to match the target server (default: 16384). The effective token maximum uses an explicit request limit or the schema's `max_tokens` default when omitted, capped by the remaining context window after tokenization. Configure `max-model-len` and the tokenizer to match the target model. Explicit `max_tokens: null` uses the remaining context window instead of the omitted-field default.
```
mayabar: Why only VLLM_MAX_N_SEQUENCES? There are additional environment variables that are part of request validation, e.g. VLLM_MAX_N_SEQUENCES

mayabar (commented):
Thanks for the rework, and for moving to the full request schema. There are several issues in the proposed implementation that need to be resolved before this can land.

## 1. Strict mode should not depend on a user-supplied schema file

Requiring `--strict-openapi` puts the burden on the user to export and carry a snapshot, and it makes strict mode depend on an artifact the simulator cannot control. The important problem: a schema exported from a newer vLLM can be inconsistent with the validations implemented in the code here, and then strict mode silently mixes two different vLLM versions - a new field passes the schema and then hits a check written for an older release, or vice versa.

Please check a snapshot into the repository and load it from there (`go:embed`), so the schema version and the hand-written checks are versioned together and `--strict` works with no extra user setup. 

## 2. The schema covers only a small part of the validation

Schema validation gives structure and types; most of what vLLM rejects is not expressible there. `n` is one example (both `n < 1` and `n > VLLM_MAX_N_SEQUENCES`, whose default is 16384 in `vllm/envs.py`), but there are many more, for example:

- `stream_options` set while `stream` is false or absent
- `top_logprobs` set while `logprobs` is false
...

This needs a systematic pass over vLLM's request models and `SamplingParams` to cover all request fields, rather than a partial list. Please also keep the exact vLLM message text where you claim parity - for instance the `n` upper-bound message here drops vLLM's second sentence about setting `VLLM_MAX_N_SEQUENCES`.

## 3. Validation logic is spread across too many places

Right now a single request is validated in four different locations, with rules duplicated between them:

- `pkg/communication/strict_schema.go`: schema validation, `n > VLLM_MAX_N_SEQUENCES`, and `min_tokens` against the schema's `max_tokens` default
- `pkg/communication/strict_validation.go`: all sampling ranges, plus `min_tokens` against the explicit `max_tokens`
- `pkg/endpoint/request.go` (`validateTokenizedRequest`): `min_tokens` again, against the remaining context window
- `pkg/api/request.go`: `strictMinTokens` / `strictDefaultMaxTokens` fields and `SetStrictTokenLimits` / `StrictTokenLimits`, only to carry values from the first location to the third (please remove new fields from the request)

So `min_tokens` versus `max_tokens` is implemented three times with three different notions of the maximum. On top of that, `pkg/endpoint/helpers.go` (`validateRequest`) already validates `n < 1` and `max_tokens < 1`, so those conditions now exist twice with two different messages depending on the flag. Please consolidate: one place per rule, one message per rule.

## 4. Align with the new engine structure

The project has been restructured recently and now has an explicit engine concept (`pkg/engine`, with `pkg/engine/vllm` supplying the vLLM-specific flags, config validation, and transport). Strict request validation should fit into that seam rather than sit beside it.

Suggested split, worth agreeing before you rework:

- Run strict validation when the request is received, in one place, so it is not interleaved with tokenization and generation.
- Keep generic OpenAI-level checks in the common path.
- Put vLLM-specific fields in a vLLM validator owned by the engine, next to the existing `ValidateConfig` - `min_tokens` is a good example, it is a vLLM `SamplingParams` extension, not an OpenAI field.

0z5a:
@mayabar Thanks for the detailed review. I have addressed the requested changes in the latest commit (`a76a5eab`): the target vLLM schemas are now embedded in the repository, the request relationship and sampling checks have been expanded with vLLM-compatible messages, validation is centralized so each rule has a single owner, and the vLLM-specific validation now lives under `pkg/engine/vllm`. The updated checks are passing. Could you please take another look when you have a chance?

pkg/communication/http.go
```diff
@@ -274,6 +274,23 @@ func (c *Communication) handleHTTP(req endpoint.Request, respBuilder responseBui
 		return
 	}
 
+	path := string(ctx.Path())
+	if c.runtime.Config().StrictRequestValidation &&
+		(path == "/v1/chat/completions" || path == "/v1/completions") {
+		if err := validateStrictContentType(string(ctx.Request.Header.ContentType())); err != nil {
+			c.sendError(ctx, err, false)
+			return
+		}
```
mayabar: Consider a function for all headers validation, currently it will contain only one header only

pkg/engine/vllm/strict_schema.go
```diff
@@ -0,0 +1,134 @@
...
+
+func loadStrictRequestValidator() (*strictRequestValidator, error) {
+	var err error
+	maxN := int64(16384)
+	if value, exists := os.LookupEnv("VLLM_MAX_N_SEQUENCES"); exists {
+		maxN, err = strconv.ParseInt(value, 10, 64)
+		if err != nil || maxN < 1 {
+			return nil, errors.New("VLLM_MAX_N_SEQUENCES must be a positive integer")
+		}
+	}
```
mayabar: Seems that there are additional environment variables that are used by vLLM for request validation, e.g. VLLM_MAX_COMPLETION_PROMPTS.

mayabar (commented):
@0z5a PR looks good, please see some comments below

mayabar (commented):
Great, thanks!
PR is ready for merge, just waiting for conflict resolution

mayabar:
@0z5a you commit was not signed ;)

0z5a:
Thanks for catching this. I re-signed the merge commit; GitHub now verifies the updated head, and the code tree is unchanged.

PR #685, opened 2026-09-05
fix(dataset): bind table-name to the SQLite table

git-jxj:
## What does this PR do?

Bind `dataset-tool --table-name` to the SQLite table name, independently of `--output-path`.

## Why is this change needed?

The flag writes into `outputPath`, so its default changes the output directory to `llmd`. Passing `--table-name custom` after `--output-path output` redirects files to `custom`; reversing the arguments keeps the output directory but still creates the `llmd` table.

## How was this tested?

Regression coverage checks the default output directory and both flag orders against the generated SQLite file and table. `CGO_ENABLED=0 go test ./pkg/dataset -run '^TestDSToolConfigurationOutput$' -count=1` passes.

Built and ran the CLI with an empty local dataset for all three cases; verified the SQLite, JSON, and dataset-card paths and table names. The same checks fail on the upstream baseline.

`make presubmit` passes formatting, linting, and module checks, then fails on existing `golang.org/x/crypto v0.55.0` findings GO-2026-6354 and GO-2026-6355. The same vulnerability check fails on unmodified upstream `0153b22`; dependencies are unchanged. The full test suite requires the vLLM render container and was not run. The commit has a GitHub-verified signature and DCO sign-off, and `scripts/check-commits.sh upstream/main` passes.

## Checklist

## Related Issues

Found through code review; the existing dataset-tool documentation describes the expected output directory and table-name behavior.

### DIFF
--- a/pkg/dataset/ds_tool_config.go
+++ b/pkg/dataset/ds_tool_config.go
@@ -67,7 +67,7 @@ func (c *DSToolConfiguration) LoadConfig() error {
 	f.StringVar(&c.outputFile, "output-file", "inference-sim-dataset",
 		"Output file name without extension, two files will be created: <output-file>.json and <output-file>.csv")
 	f.StringVar(&c.outputPath, "output-path", "", "Output path")
-	f.StringVar(&c.outputPath, "table-name", common.DefaultDSTableName, "Table name, default is 'llmd'")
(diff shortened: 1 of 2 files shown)


mayabar (approved):
Thanks for good catch!

PR #687, opened 2026-09-06
Introduce an Engine seam for config flags, validation, and HTTP/gRPC transport

irar2:
This PR splits out vLLM's own CLI flags, configuration validation, and HTTP/gRPC route binding into `pkg/engine/vllm`, behind a small Engine interface (`pkg/engine`), so `pkg/common`/`pkg/communication`/`pkg/simulator` no longer hardcode vLLM and could support a different backend later without changing their signatures. Each of those three packages declares its own minimal consumer-side interface (`common.Engine`, `communication.Transport,` `simulator.Engine`) satisfied implicitly by `vllm.Engine`, mirroring the existing `Processor`/`Runtime` pattern. 

Metrics registration and the Configuration struct's field layout are unchanged and out of scope for this PR.


pkg/communication/communication.go
```diff
@@ -55,13 +56,25 @@ func New(logger logr.Logger, processor Processor, runtime endpoint.Runtime) *Com
...
+// encoder-only mode is enabled, the gRPC server (with the active engine's
+// service added via transport.BindGRPC).
+func (c *Communication) Start(ctx context.Context, transport Transport) error {
 	c.logger.V(logging.INFO).Info("Starting communication layer")
-	return c.start(ctx)
+	return c.start(ctx, transport)
 }
 
-func (c *Communication) start(ctx context.Context) error {
+func (c *Communication) start(ctx context.Context, transport Transport) error {
```
mayabar: redundant?

pkg/communication/grpc.go
```diff
@@ -120,9 +120,10 @@ func (c *Communication) GetServerInfo(ctx context.Context, in *pb.GetServerInfoR
 
 // startGRPC starts the gRPC server and returns the server instance and an error channel.
 // It does not handle shutdown — callers are responsible for calling server.Stop().
-func (c *Communication) startGRPC(listener net.Listener) (*grpc.Server, <-chan error) {
+// transport registers the active engine's own gRPC service on server.
+func (c *Communication) startGRPC(listener net.Listener, transport Transport) (*grpc.Server, <-chan error) {
 	server := grpc.NewServer()
-	pb.RegisterVllmEngineServer(server, c)
+	transport.BindGRPC(server, c)
```
mayabar, line 126: please open an issue for case without gRPC

pkg/engine/engine.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/spf13/pflag"
+	"google.golang.org/grpc"
+
+	"github.com/llm-d/llm-d-inference-sim/pkg/common"
+	"github.com/llm-d/llm-d-inference-sim/pkg/communication"
+)
+
+// Engine supplies one backend's own CLI flags, configuration validation, and
+// HTTP/gRPC transport surface.
+type Engine interface {
```
mayabar, line 32: consider moving to common package and remove other (small) interfaces

pkg/simulator/context.go
```diff
@@ -89,6 +89,16 @@ type SimContext struct {
...
 	mooncakeEngines     map[string]map[string]string
+	// Engine is the active engine, used by admin-config updates
+	// (ApplyConfigUpdate, below) to re-validate its own configuration fields
+	// the same way the initial configuration does. Set once before the
+	// simulator starts serving; nil is treated as "nothing to validate".
+	Engine Engine
+}
+
+// Engine validates the active engine's own configuration fields.
+type Engine interface {
```
mayabar, line 100: redundant for common.Engine?

pkg/simulator/context.go
```diff
@@ -144,6 +154,11 @@ func (s *SimContext) ApplyConfigUpdate(body []byte) error {
 	if err != nil {
 		return err
 	}
+	if s.Engine != nil {
```
mayabar, line 157: consider to remove this if

pkg/communication/send_response_test.go
```diff
@@ -36,14 +35,15 @@ import (
...
-	defer func() { os.Args = oldArgs }()
-	os.Args = []string{"cmd", "--model", common.TestModelName, "--mode", common.ModeEcho}
-
-	config, err := common.ParseCommandParamsAndLoadConfig()
-	Expect(err).NotTo(HaveOccurred())
+	config := common.NewConfig()
+	config.Model = common.TestModelName
+	config.Mode = common.ModeEcho
+	config.ServedModelNames = []string{config.Model}
+	config.DisplayModelName = config.Model
```
mayabar, line 46: please check if command line simulator initialization could be used to avoid set ServedModdeNames and DisplayName explicitly

irar2: It cannot be used because of an import cycle. But this test is not e2e test anyway.

pkg/simulator/context.go
irar2: nil here means no validation is needed, and can be called from tests with nil

pkg/simulator/context.go
irar2: It's a minimal interface to only include what's needed here

pkg/engine/engine.go
irar2: Can't do that because of an import cycle

pkg/communication/grpc.go
irar2: done

pkg/communication/communication.go
irar2: yes, removed

mayabar (approved):

PR #688, opened 2026-09-06
test(metrics): replace fixed sleeps with bounded polling

### LINKED ISSUE #686: Consider using gomega.Eventually in tests in order to reduce test suite run time
Following on the idea of @0z5a in #681 to use gomega.Eventually while waiting for correct metrics, we should consider applying it for other tests and reduce the test suite run time.

0z5a:
## What does this PR do?

Replace fixed metric-readiness sleeps in existing HTTP/gRPC metrics specs with bounded `Eventually` polling (2-second timeout, 25-millisecond interval). Preserve metric assertions, simulator arguments, request spacing, and cancellation behavior.

The diff contains only `pkg/tests/metrics_test.go` and `pkg/tests/grpc_metrics_test.go`. The gRPC KV-cache changes merged in #681 are preserved unchanged.

## Why is this change needed?

Metrics become observable asynchronously. Polling the expected values avoids unconditional readiness delays and gives assertions a bounded window to observe updates, as proposed in #686.

## How was this tested?

- Focused non-KV-cache/non-prefix-cache HTTP and gRPC metrics specs passed on Darwin arm64 with Go 1.27.1 (54.524s). A temporary Go overlay skipped the unused Qwen renderer startup; selected specs, simulator arguments, and assertions were unchanged. The overlay is not part of this PR. Docker is unavailable locally, so renderer-dependent cases and the full suite were not rerun.
- `go vet ./pkg/tests` passed.
- Compared against the original metrics commit: HTTP metrics source is identical; gRPC metrics source differs only in the #681 block, which exactly matches `main`.

- With Go 1.26.6, signature/DCO checks, dependency tidiness, formatting, and lint (0 issues) passed. `make presubmit` could not finish because its `go install ...@latest` scanner step timed out reaching proxy.golang.org.
- Running the installed govulncheck v1.7.0 directly with Go 1.26.6 reports the existing x/crypto v0.55.0 findings GO-2026-6354 and GO-2026-6355. Their dependency fix is separated from this test-only PR.

## Checklist

## Related Issues

Fixes #686
Related to #681

Independent drafts: #691 (x/crypto update) and #692 (image-pinning/OpenPGP policy proposal).

### DIFF
--- a/pkg/tests/grpc_metrics_test.go
+++ b/pkg/tests/grpc_metrics_test.go
@@ -67,16 +67,18 @@ var _ = Describe("gRPC Metrics", Ordered, func() {
 			}()
 		}
 
-		time.Sleep(300 * time.Millisecond)
-		metricsResp, err := httpClient.Get(metricsUrl)
-		Expect(err).NotTo(HaveOccurred())
-		Expect(metricsResp.StatusCode).To(Equal(http.StatusOK))
+		Eventually(func(g Gomega) {
+			metricsResp, err := httpClient.Get(metricsUrl)
+			g.Expect(err).NotTo(HaveOccurred())
+			defer func() { _ = metricsResp.Body.Close() }()
+			g.Expect(metricsResp.StatusCode).To(Equal(http.StatusOK))
 
-		data, err := io.ReadAll(metricsResp.Body)
-		Expect(err).NotTo(HaveOccurred())
-		metrics := string(data)
-		Expect(metrics).To(ContainSubstring(getCountMetricLine(common.TestModelName, simulator.ReqRunningMetricName, 2)))
-		Expect(metrics).To(ContainSubstring(getCountMetricLine(common.TestModelName, simulator.ReqWaitingMetricName, 1)))
+			data, err := io.ReadAll(metricsResp.Body)
+			g.Expect(err).NotTo(HaveOccurred())
+			metrics := string(data)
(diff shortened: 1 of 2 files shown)


0z5a:
Hi @irar2 Would you mind having a review?

irar2:
@0z5a Please sign the commits

irar2:
Thanks for tackling #686 — the `Eventually` conversion itself is solid.

Could we split this into separate PRs though? Right now it bundles three unrelated changes: the flaky-test fix (#686), an x/crypto bump, and manifest image-pinning + an OpenPGP CI guard. Only the first has a linked issue. The pinning change in particular is a maintainer policy call worth its own discussion — these are example/quickstart manifests, and pinning to v0.11.2@sha256:... means they'll drift from the docs/README on every release unless something automates the bump. Also the digest should be independently verified against the registry before merge.

0z5a:
> Thanks for tackling #686 — the `Eventually` conversion itself is solid.
> 
> 
> 
> Could we split this into separate PRs though? Right now it bundles three unrelated changes: the flaky-test fix (#686), an x/crypto bump, and manifest image-pinning + an OpenPGP CI guard. Only the first has a linked issue. The pinning change in particular is a maintainer policy call worth its own discussion — these are example/quickstart manifests, and pinning to v0.11.2@sha256:... means they'll drift from the docs/README on every release unless something automates the bump. Also the digest should be independently verified against the registry before merge.

I see. Will edit soon.

PR #692, opened 2026-09-08
chore: use official release images in examples

0z5a:
## What does this PR do?

Use v0.11.2 release tags for the simulator examples and ghcr.io/llm-d/zmq-listener:v0.11.2 for the listener. Make the existing latest-tag check fail presubmit.

## Why is this change needed?

Examples should reference published releases and the official listener image. Version tags keep the manifests readable and avoid introducing a digest-update policy. The existing govulncheck remains responsible for vulnerability checks; no OpenPGP import guard is included.

## How was this tested?

- make check-latest-tags passed.
- git diff --check passed.
- No Go runtime code changed; no deployment was performed.

## Related Issues

Related to #688. Release-update policy remains under maintainer discussion; this PR remains a draft.

### DIFF
--- a/manifests/disaggregation/vllm-sim-pd.yaml
+++ b/manifests/disaggregation/vllm-sim-pd.yaml
@@ -47,7 +47,7 @@ spec:
           periodSeconds: 5
       containers:
         - name: vllm-prefill
-          image: ghcr.io/llm-d/llm-d-inference-sim:latest
+          image: ghcr.io/llm-d/llm-d-inference-sim:v0.11.2@sha256:32144df791330a0006b747edfdf2b114a0fe728e023a9d1b3463eeb48d32abb9
           imagePullPolicy: IfNotPresent
           args:
             - "--v=4"
@@ -119,7 +119,7 @@ spec:
         ports:
           - containerPort: 8000
       - name: vllm-decode
-        image: ghcr.io/llm-d/llm-d-inference-sim:latest
+        image: ghcr.io/llm-d/llm-d-inference-sim:v0.11.2@sha256:32144df791330a0006b747edfdf2b114a0fe728e023a9d1b3463eeb48d32abb9
         imagePullPolicy: IfNotPresent
         args:
           - "--v=4"
--- a/manifests/zmq-listener/deploy_listener.yaml
+++ b/manifests/zmq-listener/deploy_listener.yaml
@@ -31,7 +31,7 @@ spec:
     spec:
       containers:
         - name: listener
-          image: quay.io/mayab/zmq-listener:latest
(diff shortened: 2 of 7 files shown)


manifests/disaggregation/vllm-sim-pd.yaml
```diff
@@ -47,7 +47,7 @@ spec:
           periodSeconds: 5
       containers:
         - name: vllm-prefill
-          image: ghcr.io/llm-d/llm-d-inference-sim:latest
+          image: ghcr.io/llm-d/llm-d-inference-sim:v0.11.2@sha256:32144df791330a0006b747edfdf2b114a0fe728e023a9d1b3463eeb48d32abb9
```
mayabar: Why digests is required here?  Why version label is not enough?

manifests/zmq-listener/deploy_listener.yaml
```diff
@@ -31,7 +31,7 @@ spec:
     spec:
       containers:
         - name: listener
-          image: quay.io/mayab/zmq-listener:latest
+          image: quay.io/mayab/zmq-listener@sha256:2427928648d25b1195dc2e226304d3774974058b8f3f6d45099b07d4814430b6
```
mayabar: `quay.io/mayab/zmq-listener` is a personal namespace, and that image is an old local experiment. It predates this PR, so it isn't something you got wrong, but since this PR is already changing that code, let's switch to the official image the release pipeline publishes: `ghcr.io/llm-d/zmq-listener` with tag `v0.11.2`

scripts/check-no-openpgp.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
mayabar: We don't import golang.org/x/crypto/openpgp, so this check has nothing to catch — and govulncheck already covers the packages we do import.
Please remove this script

mayabar (commented):
Some comments on this PR

manifests/disaggregation/vllm-sim-pd.yaml
0z5a: A digest guarantees immutable contents if a tag is moved, but that extra constraint is not needed for these examples. I removed the digest pins and kept the v0.11.2 release tags in 34a058a.

manifests/zmq-listener/deploy_listener.yaml
0z5a: Changed the listener to ghcr.io/llm-d/zmq-listener:v0.11.2 in 34a058a. Thanks for pointing me to the official release image.

scripts/check-no-openpgp.sh
0z5a: Removed the script, Makefile target, and presubmit dependency in 34a058a. The existing govulncheck remains in place. make check-latest-tags and git diff --check passed.

mayabar (approved):

PR #693, opened 2026-09-08
Rework metrics into a unified bus/adapter shared by real and fake metrics

mayabar:
## What does this PR do?

Replaces the simulator's scattered metrics-writing code with a single metrics bus and vLLM metrics adapter (`pkg/metrics`), so real request processing and fake-metrics generation update Prometheus state through the same functions instead of duplicated paths. Removes the old metrics implementation embedded in `pkg/simulator`. Consolidates LoRA metric updates onto a single channel and fixes a related race condition.

## Why is this change needed?

Metrics logic was tied directly into the simulator's vLLM-specific code paths, with real and fake metrics writing through separate, duplicated logic. Routing both through a single metrics bus and adapter decouples metrics from the vLLM engine specifics and gives the two a shared write path, so the simulator can add support for other engines (e.g. SGLang) without duplicating metrics logic per engine.


mayabar:
Closing this PR in favor of PR #714

PR #696, opened 2026-09-09
Group KV-cache settings into a KVCacheConfig struct

irar2:
Pulls the KV-cache sizing/hashing/eventing settings out of the flat `Configuration` struct into a new `KVCacheConfig` sub-struct, nested under a `kvcache:` key in YAML (old flat top-level keys still work for backward compatibility) while CLI flag names stay unchanged.

KV-cache transfer latencies and global-cache-hit-threshold remain on Configuration as before. 

When `enable-kvcache` is false, the whole KVCache block now reports as all-zero rather than leaking unused defaults or inert flag/YAML values.


pkg/common/config.go
```diff
@@ -412,13 +426,47 @@ func (c *Configuration) load(configFile string) error {
...
+	}
+	for _, key := range kvCacheYAMLKeys {
+		if v, ok := raw[key]; ok {
+			if _, exists := nested[key]; !exists {
+				nested[key] = v
+			}
+			delete(raw, key)
+		}
+	}
+	if len(nested) > 0 {
```
mayabar, line 475: Please take care of case when part of kv params in the root and part is nested - return error

pkg/common/config_test.go
```diff
@@ -268,3 +270,63 @@ var _ = Describe("admin struct tags", func() {
...
+model: test-model
+kv-cache-size: 111
+kvcache:
+  kv-cache-size: 222
+`))).To(Succeed())
+
+		Expect(c.KVCache.KVCacheSize).To(Equal(222))
+	})
+
+	It("leaves global-cache-hit-threshold untouched by kv-cache folding", func() {
```
mayabar: do we need this test?

mayabar (dismissed):

mayabar (approved):

PR #699, opened 2026-09-10
fix(responses): include instructions in prompt tokenization

### LINKED ISSUE #698: Responses instructions are omitted from prompt tokenization and input usage
### What happened?

On main (`24914d3`), `/v1/responses` echoes the `instructions` field but omits it from the messages passed to the tokenizer. Input usage and the prompt used for simulated prefill/KV-cache lookup therefore omit the system instruction.

### Reproduction

Build and start the simulator with its built-in tokenizer:

```sh
go build -o /tmp/inference-sim ./cmd/llm-d-inference-sim
/tmp/inference-sim --model test --mode echo --port 8000
```

Send these requests to `/v1/responses`:

```json
{"model":"test","input":"Hello world","max_output_tokens":32}
```

```json
{"model":"test","input":"Hello world","instructions":"Reply in French","max_output_tokens":32}
```

Both return `usage.input_tokens: 7`. The equivalent `/v1/chat/completions` request returns `usage.prompt_tokens: 15`:

```json
{"model":"test","max_tokens":32,"messages":[{"role":"system","content":"Reply in French"},{"role":"user","content":"Hello world"}]}
```

### Expected behavior

Instructions should be tokenized as a system message before the input, matching vLLM's Responses input construction. With the built-in tokenizer, the instructed Responses request should therefore report 15 input tokens, matching the equivalent chat request.

### Cause

`responsesReqCtx.encode()` calls `convertInputToMessages(r.req.Input)` and renders the result without adding `r.req.Instructions`.

Reproduced locally with the compiled server and actual HTTP requests, and with a failing endpoint regression test. A six-line change prepending the system message makes the HTTP comparison pass (7 without instructions, 15 with instructions, 15 for chat).

git-jxj:
## What does this PR do?

Prepend `instructions` as a system message when tokenizing `/v1/responses` input. This makes input usage and the prompt supplied to prefill/cache simulation include the instruction, matching equivalent chat input.

## Why is this change needed?

A request with `input: "Hello world"` and `instructions: "Reply in French"` reports 7 input tokens with the built-in tokenizer, just like the request without instructions. The equivalent chat request reports 15. With this fix the instructed Responses request also reports 15; the uninstructed request still reports 7.

## How was this tested?

The new regression coverage checks token IDs and token strings for input with and without instructions, including system-message ordering and preservation of the original input.

`CGO_ENABLED=0 go test ./pkg/endpoint` passes all 54 specs. The instructed case fails on the original code.

Built both original and fixed server binaries and sent real HTTP requests to `/v1/responses` and `/v1/chat/completions`. The input-token comparison fails before the fix (7 vs 15) and passes after it (15 vs 15). This uses the built-in tokenizer; no model server is required.

`make presubmit` passes signing, module tidiness, formatting, and lint (0 issues), then fails on existing `golang.org/x/crypto@v0.55.0` findings in the testcontainers SSH path. An independent scan of unchanged upstream `24914d3` reports the same findings. Dependency update #691 already addresses that module; this PR does not duplicate it.

## Checklist

The complete Docker-backed `make test` suite was not run; the affected endpoint package and real HTTP comparison were run as described above.

## Related Issues

Fixes #698


irar2:
@git-jxj Please fix the test that now fails because instructions are also cached

git-jxj:
@irar2 Thanks for catching this. Fixed in 5d9f4c3. Each rendered prompt fills three blocks; the French requests share two prefix blocks, and the English requests share two separate prefix blocks. The four requests therefore occupy eight blocks in total, so the expected usage is 8/16 = 0.5.

Reproduced the failure with the old assertion and verified the updated test passes using vllm/vllm-openai-cpu:v0.21.0. The full `make test` run passes all 10 suites (880 specs). Locally, the render container was started with nerdctl, with a temporary Go overlay connecting the test tokenizer manager to that container. Lint passes; presubmit still stops at the already documented x/crypto vulnerability findings.

PR #701, opened 2026-09-10
Make prefix cache hit rate match vLLM definition

Lucas-Fernandes-Martins:
## What does this PR do?

Hi, I believe the current way prefix blocks are counted doesn't account for evictions. 

```
                } else if _, exists := bc.usedBlocks[bKey]; !exists {
			// new block — record its index so tokens can be written after
			// the capacity check passes, preventing orphaned entries on error.
			blocksToAdd = append(blocksToAdd, newBlock{key: bKey, tokenIdx: i})
		} else {
			blockAlreadyInUse = append(blockAlreadyInUse, bKey)
			lastCachedIdx = i
```

Here, even after identifying a block that is not in the prefix, we keep updating `lastCachedIdx`. However, in my view it is possible for one of these blocks to have been evicted.

Example:

We send a request with blocks [1,2,3] which get cached.

Then block 1 gets evicted.

If we send [1,2,3] again, the current logic will state 2 blocks have been found in the cache, when the actual useful prefix is 0.

I provide a very simple fix which involves stop incrementing `lastCachedIdx` once we find a block that is not in the cache.

## Why is this change needed?

To make sure prefix hits are computed correctly.

## How was this tested?

Added a new regression test.

- [x ] Unit tests added/updated
- [ ] Manual testing performed

## Checklist

- [x ] Commits are signed off (`git commit -s`) per [DCO](https://github.com/llm-d/.github/blob/main/PR_SIGNOFF.md)
- [x ] Code follows project [contributing guidelines](https://github.com/llm-d/.github/blob/main/CONTRIBUTING.md)
- [ x] Tests pass locally (`make test`)
- [x ] Linters pass (`make lint`)

## Related Issues

#705


pkg/kvcache/block_cache.go
```diff
@@ -202,18 +202,27 @@ func (bc *blockCache) startRequest(req Request, blockHashes []uint64, blockToken
 	// as a contiguous leading prefix — the parent of the first new block is
 	// always blockHashes[lastCachedIdx].
```
mayabar, line 203: please update the comment according the fix in the code

pkg/kvcache/block_cache.go
mayabar, line 160: please update the comment according the new behavior

pkg/kvcache/block_cache.go
```diff
@@ -202,18 +202,27 @@ func (bc *blockCache) startRequest(req Request, blockHashes []uint64, blockToken
...
 	lastCachedIdx := -1
+	cachedPrefixBlocks := 0
```
mayabar, line 205: cachedPrefixBlocks is always lastCachedIdx+1, consider to remove

mayabar (commented):
@Lucas-Fernandes-Martins Thanks!

Currently existing test sets `unusedBlocks/blockToTokens` by hand, so it checks the counting rule but never creates a hole the way real code does. Going through start/finish with a forced eviction would also cover the parent-hash behavior and catch regressions.

The eviction algorithm itself needs updating and I've opened #704 for it. The simulator can evict a block from the start or middle of a chain and keep the ones after it, which vLLM never does, so hit rates and simulated TTFT diverge non-deterministically.

PR #707, opened 2026-09-14
Move engine-specific config construction and validation into pkg/engine/vllm

irar2:
Fields the simulator reads at runtime stay on `common.Configuration`, but the
CLI/YAML parsing that builds the engine-specific ones now lives with the engine
that owns their wire format: `Engine.BindFlags` takes the raw YAML tree and folds
and unmarshals its own "lora", "kvcache", and "fake-metrics" blocks, so
`common.Configuration.load` no longer needs to know vLLM's layout.

`Configuration.FakeMetrics` becomes an interface implemented by
`pkg/engine/vllm/fakemetrics`, since the metrics an engine can fake mirror its own
metric surface. 

Five vLLM CLI flags that nothing ever read back
(`--mm-processor-kwargs`, `--ec-transfer-config`, `--enforce-eager`,
`--enable-prefix-caching`, `--tensor-parallel-size`) bind to locals in the engine
instead of `Configuration` fields. They still parse.


pkg/engine/vllm/flags.go
```diff
@@ -22,15 +22,114 @@ import (
...
+	var yamlLoraModules []string
+	if rawYAML != nil {
+		if err := common.FoldLegacyKeys(rawYAML, "lora", loraLegacyFlatKeys); err != nil {
+			return err
+		}
+		// Seeded from the current values so that keys the block omits keep
+		// their defaults while an explicit zero still comes through (and is
+		// then rejected by ValidateConfig, rather than silently ignored).
+		// loraYAML is not convertible from common.LoraConfig the way
+		// kvCacheYAML is, since LoraModules differs in type.
```
mayabar: is it possible that cfg.Lora is nil?

pkg/engine/vllm/flags.go
```diff
@@ -22,15 +22,114 @@ import (
...
+		}
+		cfg.Lora.MaxLoras = ly.MaxLoras
+		cfg.Lora.MaxCPULoras = ly.MaxCPULoras
+		yamlLoraModules = ly.LoraModules
+	}
+
+	// A YAML config file's "kvcache" block uses vLLM's own wire format,
+	// folded and unmarshaled here (rather than by common.Configuration.load)
+	// since a different engine could shape KV-cache config differently.
+	if rawYAML != nil {
```
mayabar: why not to use the "if" above?

pkg/engine/vllm/flags.go
```diff
@@ -59,24 +158,27 @@ func (Engine) BindFlags(f *pflag.FlagSet, cfg *common.Configuration) error {
...
-	f.StringVar(&cfg.ECTransferConfig, "ec-transfer-config", cfg.ECTransferConfig, "Configuration for distributed EC cache transfer, ignored")
-	common.AddToggle(f, &cfg.EnforceEager,
+
+	// These vLLM CLI flags are accepted for command-line compatibility and
+	// ignored: nothing in the simulator ever reads them back, so unlike the
+	// rest of this engine's flags they have no Configuration field to bind
+	// to, just a local variable to satisfy pflag's API.
+	var mmProcessorKWArgs, ecTransferConfig string
+	var enforceEager, enablePrefixCaching bool
+	var tpSize int
```
mayabar: consider using dummyIt/Str/... var names

pkg/engine/vllm/flags.go
```diff
@@ -86,10 +188,11 @@ func (Engine) BindFlags(f *pflag.FlagSet, cfg *common.Configuration) error {
 	f.Lookup("fake-metrics").NoOptDefVal = dummy
 	f.Lookup("fake-metrics").DefValue = ""
 
-	if configFileGiven {
-		if err := unmarshalLoraFakeMetrics(cfg); err != nil {
+	if yamlFakeMetrics != nil {
```
mayabar: consider to move this section to the place where lora fake metrics are loaded from the yaml

pkg/engine/vllm/flags.go
```diff
@@ -100,14 +203,13 @@ func (Engine) BindFlags(f *pflag.FlagSet, cfg *common.Configuration) error {
 		}
 	}
 
-	if configFileGiven {
-		if err := unmarshalLoras(cfg); err != nil {
+	if rawYAML != nil {
+		if err := unmarshalLoras(cfg, yamlLoraModules); err != nil {
```
mayabar: same

pkg/engine/vllm/flags.go
irar2: No, it's not a pointer, it's a struct:
Lora LoraConfig `yaml:"-" json:"lora"`

mayabar (approved):

PR #720, opened 2026-09-17
Gate gRPC server startup on engine gRPC support

### LINKED ISSUE #690: Support correct gRPC server creation for different engines

irar2:
`Communication.Start` always opened an HTTP/2-matched listener and spun up a `grpc.Server` regardless of what the active engine actually supports, and `Transport.BindGRPC`'s bool return (meant to signal "no gRPC surface") was discarded at the call site. This wires that bool through: `bindGRPC` now runs before any cmux matching, and the HTTP/2 listener and gRPC serve loop are only created when the engine reports gRPC support. 

Fixes #690


pkg/communication/communication.go
```diff
@@ -84,16 +86,19 @@ func (c *Communication) Start(ctx context.Context, transport Transport) error {
...
-		// gRPC uses HTTP/2
-		grpcL := m.Match(cmux.HTTP2())
-		grpcServer, grpcErrCh = c.startGRPC(grpcL, transport)
-		// Check for an immediate startup error.
-		select {
-		case err := <-grpcErrCh:
-			if err != nil {
-				return err
+		if server, ok := c.bindGRPC(transport); ok {
+			grpcServer = server
```
mayabar, line 90: why grpcServer var is required?

irar2: It is used in line 170 to shut it down

mayabar (approved):

PR #724, opened 2026-09-20
Add stateless /v1/responses/render endpoint

### LINKED ISSUE #708: Add /v1/responses/render endpoin
Recently, vLLM was extended to support the stateless `/v1/responses/render endpoint [#50195](https://github.com/vllm-project/vllm/pull/50195).
Could we add the same functionality to the simulator as well?

804533125:
## What does this PR do?

Adds a stateless `POST /v1/responses/render` endpoint to the simulator, mirroring the one vLLM added in vllm-project/vllm#50195. The simulator already had the render endpoints for chat completions and text completions, so this just fills in the Responses half — handy for anyone exercising the `render` -> `generate` flow against the sim, and it keeps our OpenAI-compatible surface in step with upstream vLLM.

## Why is this needed?

vLLM's `/v1/responses/render` exists for disaggregated / scale-out frontends: a GPU-less renderer applies the model-specific formatting and tokenization and returns the rendered prompt for a separate token-in generation service to run. Without the Responses variant, the sim couldn't stand in for that half of the pipeline.

## What changed

- `ResponsesRequest` now implements `RenderableRequest`, so the handler is a single line that reuses the existing render path (parse -> validate -> render -> response).
- The one bit worth calling out: the input-to-messages conversion plus the `instructions` handling used to live only inside the worker-pipeline `encode()`. I pulled that into a shared `buildResponsesMessages` so the `/v1/responses` generation path and the new render path build the prompt the *same* way — that's what makes the render output trustworthy.
- The response is a single `RenderResponse` object (`token_ids` + `features` for multimodal input), matching the other render endpoints rather than vLLM's fuller `GenerateRequest` shape. That's a deliberate call to stay consistent with the sim's existing house style — happy to match vLLM's wire shape instead if you'd prefer.
- `previous_response_id` is rejected with `400 Bad Request`, because the render endpoint is stateless — the caller is expected to rehydrate response state before calling it, same as upstream.

## How was this tested?

New unit tests cover render parity (the render endpoint produces the same token IDs as the generation path, with/without instructions, multi-message input) and the validation boundary (`previous_response_id` rejected, empty `input` rejected). New integration entries cover string input, message-array input with instructions, `input_image` multimodal, the proxy mode, and the two 400 cases.

On my Windows machine I went a step further than the suite and drove the actual binary end to end:

- simulated mode: all six scenarios above return the expected `token_ids` / `features`, with 400s on the invalid bodies.
- proxy mode: ran a real vLLM CPU render container with `Qwen/Qwen2-VL-2B-Instruct` and pointed the sim at it. A text input came back with real Qwen token IDs (151644 ...), and an `input_image` input came back with the upstream Qwen2-VL multimodal `features` — so the local responses-to-messages conversion and the upstream forwarding both work.

One environment note: the `pkg/tests` suite itself can't run on this box — testcontainers v0.44.0's `WithWaitStrategy` overrides the HTTP wait strategy's own `WithStartupTimeout` with a hard 60s deadline, and re-downloading the model into the ephemeral tmpfs `/.cache` doesn't fit in that window. The code path itself is verified by the manual runs above, but please let CI (`make presubmit`, Linux) be the word on the full suite.

## Checklist

## Related Issues

Fixes #708


804533125:
@mayabar  hi，ptal,thanks!

pkg/endpoint/responses.go
```diff
@@ -54,6 +55,35 @@ func (r *ResponsesRequest) Validate(toolsValidator *ToolsValidator) *api.Error {
...
 
+// ValidateBody checks that a responses body has a non-empty input — the
+// minimum shape required for /v1/responses/render. The stateless render
+// boundary rejects previous_response_id, mirroring vLLM: a stateful request
+// must be rehydrated by the caller before it can be rendered.
+func (r *ResponsesRequest) ValidateBody() *api.Error {
+	if r.PreviousResponseID != nil {
+		serverErr := api.NewError("previous_response_id is not supported on /v1/responses/render",
+			fasthttp.StatusBadRequest, nil)
+		return &serverErr
```
irar2, line 66: ValidateBody should check that the unmarshalled body matches the endpoint's expected shape. Your implementation puts an endpoint-specific policy, reject previous_response_id, into a method on ResponsesRequest, the same type /v1/responses uses. Anyone later wiring Validate to ValidateBody for symmetry with the other two endpoints (completions and chat completions) would silently start 400-ing a field on /v1/responses. The shape check and the stateless-boundary rejection should not be the same method.

docs/http-endpoints.md
```diff
@@ -488,6 +502,8 @@ For everything else, behavior depends on the active tokenizer (selected automati
 - **HuggingFace tokenizer** (real model): each text prompt and chat-completions request is forwarded to the upstream vLLM render service at `--render-url`. For chat requests, `mm_features` returned by the upstream are passed through.
 - **Simulated tokenizer** (dummy model): the simulator tokenizes locally using its regex-based splitter. For chat requests containing `image_url`, `audio_url`, `input_audio`, or `video_url` blocks, synthetic `mm_features` are produced so multimodal-aware downstream code paths can be exercised without a real renderer.
 
+`/v1/responses/render` reuses the same prompt rendering as `/v1/responses`: `input` items are converted to messages, `instructions` is prepended as a system message, and the result is tokenized. It returns a single `token_ids`/`features` object rather than an array. Being stateless, it rejects `previous_response_id` with `400 Bad Request`.
+
```
irar2, line 506: I think it should be explained here that the sim converts input to chat messages locally and asks the upstream to render a chat request. vLLM's render_responses is a distinct path with responses-specific formatting, so the token IDs will not match a real /v1/responses/render for the same body. This matches what the sim already does on the /v1/responses generation path, so it is consistent - but the docs should not imply the output mirrors vLLM's.

irar2 (commented):
The comment in pkg/endpoint/request.go:47-50 needs to be updated to include responses/render.

The `RenderResponse` versus vLLM's `GenerateRequest` shape: it's consistent with the rest of the endpoints, so it's OK for this PR. But I would like to open an issue to change the response for all the render endpoints in the simulator.

